PR #2, opened 2016-02-28
YapDatabase 2.8.2: update to conform to new YapDatabaseSecondaryIndex…

isghe:
…WithObjectBlock
https://github.com/yapstudios/YapDatabase/wiki/Changelog

> 2.8.2 (2016-2-3)
> 
> API Change: YapDatabaseSecondaryIndexHandler now has a transaction parameter.

In YapDatabase commit 44f52f300d1ddb066c674152b0e4baa2fd98020f the
typedef of YapDatabaseSecondaryIndexWithObjectBlock changed, adding
the parameter YapDatabaseReadTransaction *transaction.

Because of that TSDatabaseSecondaryIndexes.m failed compiling.
Adding the new (unused) arg "YapDatabaseReadTransaction *transaction",
make it compilabile again.

I didn't test on runtime.

PR #3, opened 2016-03-11
Delete last message updates thread preview (signal ios#833)

### LINKED ISSUE #833: Deleting the last message does not update conversation list snippet
1. Open a conversation that has more than one message
1. Delete the last message from that conversation
1. Return back to conversation list
1. The message snippet for that conversation is empty

iPhone 4S
iOS 8.3
Signal 2.1.1

(edit: modified the steps to be more explicit)

michaelkirk:
This fixes https://github.com/WhisperSystems/Signal-iOS/issues/833

Note that I had to merge #2 locally before I was able to build this.

img_0314

And I had to merge https://github.com/WhisperSystems/Signal-iOS/pull/1102 in order to delete messages to test this.

### DIFF
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -19,7 +19,6 @@
 @property (nonatomic, retain) NSDate *creationDate;
 @property (nonatomic, copy) NSDate *archivalDate;
 @property (nonatomic, retain) NSDate *lastMessageDate;
-@property (nonatomic, copy) NSString *latestMessageId;
 @property (nonatomic, copy) NSString *messageDraft;
 
 - (TSInteraction *) lastInteraction;
@@ -37,7 +36,6 @@
 
     if (self) {
         _archivalDate    = nil;
-        _latestMessageId = nil;
         _lastMessageDate = nil;
         _creationDate    = [NSDate date];
         _messageDraft    = nil;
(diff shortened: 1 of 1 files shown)


src/Contacts/TSThread.m
```diff
@@ -19,8 +19,10 @@ @interface TSThread ()
 @property (nonatomic, retain) NSDate *creationDate;
 @property (nonatomic, copy) NSDate *archivalDate;
 @property (nonatomic, retain) NSDate *lastMessageDate;
-@property (nonatomic, copy) NSString *latestMessageId;
 @property (nonatomic, copy) NSString *messageDraft;
+
+- (TSInteraction *) lastInteraction;
+
 @end
```
FredericJacobs, line 26: Why would you store another serialization of that interaction? I don't see why you would duplicate it in the database in the thread.

michaelkirk, line 26: I'm not storing another serialization of the interaction - take a peak at the implementation a few lines down.

Essentially, I'm just taking advantage of our existing YapDatabase view which _already_ keeps track of lastObjectInGroup, vs us redundantly (and incorrectly) reimplementing some of that logic via our "lastMessageId" field.

edit: The reason this fixes (https://github.com/WhisperSystems/Signal-iOS/issues/833) is that we don't have to worry about updating the "lastMessageId" when the last message is deleted. Yap's already tracking it for us.

FredericJacobs, line 26: My bad. Read the GitHub diff to quickly.

michaelkirk, line 26: No worries. Just happy to see you in the PRs. =)

PR #19, opened 2016-07-29
Handle animated content type in TSAttachment:descritpion:

ArkadiuszBanas:
Displaying icons for gifs was missing in messages list cells
(https://github.com/WhisperSystems/Signal-iOS/issues/1271).

//FREEBE


michaelkirk:
Great. Thank you!

PR #22, opened 2016-08-06
Respect user display name order

abcdev:
return contact full name depending on user system-wide preference.


src/Contacts/Contact.m
```diff
@@ -37,12 +37,17 @@ - (instancetype)initWithContactWithFirstName:(NSString *)firstName
 
 - (NSString *)fullName {
     NSMutableString *fullName = [NSMutableString string];
-    if (self.firstName)
+    if (self.firstName) {
         [fullName appendString:self.firstName];
+        }
     if (self.lastName) {
```
michaelkirk, line 43: nit: indent on line #42

michaelkirk:
Thanks for working on this.

Since we're now trimming, I think this whole thing would be a little more clear as just:

```
 - (NSString *)fullName
{

    NSString *fullstring;
    if (ABPersonGetCompositeNameFormat() == kABPersonCompositeNameFormatFirstNameFirst) {
        fullName = [NString stringWithFormat:"%@ %@", firstName, lastName];
    } else {
        fullName = [NString stringWithFormat:"%@ %@", lastName, firstName];
    }

    // Trim whitespace from missing name components
    return [fullName stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
}
```

abcdev:
If one of the name components is nil you will end up with "(null)" in the string. Making stringByTrimmingCharactersInSet useless. As far as I understood if there is anything contact name related but not first name it ends up being in the lastName signal contact hence the leading "if (self.firstName)"

roostr:
This work was folded into the now merged #67, so this PR is no longer valid. Thanks for the inspiration on a clean way to do the formatting, @abcdev.

michaelkirk:
Thanks @roostr, and sorry to let yours go stale @abcdev.

PR #47, opened 2016-10-15
Maps numbers to names in TSGroupModel

mdouglass:
# Use an injected ContactsManager to look up proper user names for phone numbers in [TSGroupModel getInfoStringAboutUpdateTo].

Fixes WhisperSystems/Signal-iOS#757

// FREEBIE

### DIFF
--- a/src/Messages/TSGroupModel.h
+++ b/src/Messages/TSGroupModel.h
@@ -2,6 +2,7 @@
 //  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
 
 #import "TSYapDatabaseObject.h"
+#import "ContactsManagerProtocol.h"
 
 @interface TSGroupModel : TSYapDatabaseObject
 
@@ -19,7 +20,7 @@
 
 - (BOOL)isEqual:(id)other;
 - (BOOL)isEqualToGroupModel:(TSGroupModel *)model;
-- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)model;
+- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)model withContactsManager:(id<ContactsManagerProtocol>)contactsManager;
 #endif
 
 
--- a/src/Messages/TSGroupModel.h
+++ b/src/Messages/TSGroupModel.h
@@ -20,7 +20,7 @@
 
 - (BOOL)isEqual:(id)other;
 - (BOOL)isEqualToGroupModel:(TSGroupModel *)model;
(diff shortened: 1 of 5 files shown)


mdouglass:
This is my first commit to this project, so feel free to beat up on it.  I think I've properly following the contribution guidelines and I've submitted a CLA. Let me know if I missing anything else.

src/Messages/TSGroupModel.h
```diff
@@ -19,7 +20,7 @@
 
 - (BOOL)isEqual:(id)other;
 - (BOOL)isEqualToGroupModel:(TSGroupModel *)model;
-- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)model;
+- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)model withContactsManager:(id<ContactsManagerProtocol>)contactsManager;
```
michaelkirk: style nit, name the parameter without the preposition, just `contactsManager:(id<ContactsManagerProtocol>)contactsManager`

michaelkirk (commented):
Thanks for this.

This works when someone else updates the group.

Beyond your improvement here, it would be good to address what happens when _I'm_ the one updating the group. As it is, I see just "Group Updated" in that case.

See: 
`- (void)updateGroupModelTo:(TSGroupModel *)newGroupModel` in MessagesViewController.m for a similar method.

Also, if you're into using cool functions like `map` and injecting dependencies, there's plenty more work to be done in Signal iOS. =)

mdouglass:
I'll update the parameter name.  
Should I make the other change in this PR or open another one for that issue?
I'm definitely planning to do more work on Signal-iOS.  Wanted to pick something small to get a foot in the door :)

michaelkirk:
Getting group updates made locally to show these details will likely require changes to Signal-iOS, so it will necessitate a PR over there.

I was thinking it made sense to wait on merging this one so you could make sure any shared code works in both cases.

mdouglass:
That makes sense, I'll take a look.

michaelkirk:
Here's where we're currently handling it
https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/view%20controllers/MessagesViewController.m#L2011

Seemingly as a result of:
https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/view%20controllers/MessagesViewController.m#L2041

mdouglass:
Ack, something went wrong with that merge before I committed it. I'll have to redo it.

michaelkirk:
Feel free to force push your own branches in the future.

mdouglass:
Good to know.  Never quite sure when that's ok with git.  Too much time spent in p4.

PR #48, opened 2016-10-17
Fix signal ios issue 757 take 2

mdouglass:


mdouglass:
Dependent change is at https://github.com/WhisperSystems/Signal-iOS/pull/1400.
I'm not sure how to handle rev'ing the library so that passes.

michaelkirk:
> I'm not sure how to handle rev'ing the library so that passes.

Yeah, it's a little messy. Ultimately the main thing is to see that the Signal-iOS tests pass with your changes to SSK. 

So, the best thing is to update the Podfile to point at your fork of SignalServiceKit, at the same branch as your SSK PR. Run pod install and commit those changes in your Signal-iOS PR. Then we can see that the tests pass.

```
 target 'Signal' do
     pod 'SocketRocket',               :git => 'https://github.com/facebook/SocketRocket.git'
-    pod 'SignalServiceKit',           git: 'https://github.com/WhisperSystems/SignalServiceKit.git'
+    pod 'SignalServiceKit',           git: 'https://github.com/mdouglass/SignalServiceKit.git', branch: 'my-feature-branch'
     #pod 'SignalServiceKit',           path: '../SignalServiceKit'
```

I clean up the references as part of merging. (e.g. https://github.com/WhisperSystems/Signal-iOS/commit/8d0b579a961d25bf6de630eb4491d402348f4da3). If it freaks you out that I'm editing your commit, commit Podfile changes as a separate commit, and I'll simply replace that commit wholesale.

PR #67, opened 2016-11-23
Contacts only calculate fullName once and have sortable fullnames

roostr:
This PR enables a fix of [Signal-iOS #1487](https://github.com/WhisperSystems/Signal-iOS/issues/1487)

1. Adds caching of the calculated fullName value (which will slightly improve performance)
2. Incorporates fullNames that respect the first-name-first rules of the currently unmerged PR #22
3. Adds two new fullName properties that can be used for sorting comparators

### DIFF
--- a/src/Contacts/Contact.h
+++ b/src/Contacts/Contact.h
@@ -12,12 +12,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class CNContact;
 @class PhoneNumber;
+@class UIImage;
 
 @interface Contact : NSObject
 
 @property (nullable, readonly, nonatomic) NSString *firstName;
 @property (nullable, readonly, nonatomic) NSString *lastName;
 @property (readonly, nonatomic) NSString *fullName;
+@property (readonly, nonatomic) NSString *comparableNameFirstLast;
+@property (readonly, nonatomic) NSString *comparableNameLastFirst;
 @property (readonly, nonatomic) NSArray<PhoneNumber *> *parsedPhoneNumbers;
 @property (readonly, nonatomic) NSArray<NSString *> *userTextPhoneNumbers;
 @property (readonly, nonatomic) NSArray<NSString *> *emails;
@@ -42,6 +45,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithContact:(CNContact *)contact;
 
 #endif // TARGET_OS_IOS
+
++ (NSComparator)comparatorSortingNamesByFirstThenLast:(BOOL)firstNameOrdering;
 
 @end
 
--- a/src/Contacts/Contact.m
+++ b/src/Contacts/Contact.m
@@ -9,6 +9,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation Contact
 
+@synthesize fullName = _fullName;
+@synthesize comparableNameFirstLast = _comparableNameFirstLast;
+@synthesize comparableNameLastFirst = _comparableNameLastFirst;
+
 #if TARGET_OS_IOS
 - (instancetype)initWithContactWithFirstName:(nullable NSString *)firstName
                                  andLastName:(nullable NSString *)lastName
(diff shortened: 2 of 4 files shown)


src/Contacts/Contact.m
```diff
@@ -36,17 +50,54 @@ - (instancetype)initWithContactWithFirstName:(NSString *)firstName
...
+            _fullName = [self combineLeftName:_firstName withRightName:_lastName usingSeparator:@" "];
+        } else {
+            _fullName = [self combineLeftName:_lastName withRightName:_firstName usingSeparator:@" "];
+        }
+    }
+    
+    return _fullName;
+}
+
+- (NSString *)comparableNameFirstLast {
```
michaelkirk, line 109: Nice, can you add some test cases for the sorting scenarios in your comment? 

SSK tests are run in the context of ../SignalServiceKit/Example/TSKitiOSTestApp/

src/Contacts/Contact.m
```diff
@@ -3,6 +3,15 @@
 #import "SignalRecipient.h"
 #import "TSStorageManager.h"
 
+@interface Contact ()
+{
```
michaelkirk: How about using @synthesize instead?

src/Contacts/Contact.m
```diff
@@ -19,6 +28,11 @@ - (instancetype)initWithContactWithFirstName:(NSString *)firstName
         _recordID             = record;
         _image                = image;
 
+        // Next three fields are generated on-demand on first access
+        _fullName                = nil;
```
michaelkirk: not necessary to initialize ivars

michaelkirk (commented):
Glad to see this. Just some minor feedback

src/Contacts/Contact.m
roostr: Yup, sure thing.

src/Contacts/Contact.m
roostr: True. Included this to be declarative and reduce chance of ambiguity for future readers/contributors, but I'm fine with removing it in the interest of brevity.

src/Contacts/Contact.m
roostr: Absolutely!

roostr:
All feedback has been incorporated in the new 2c1fc80 branch. Also, moved the comparator into the Contact model object for ease of testing. You'll notice that Contact.h now includes UIKit, that's to resolve a compiling issue on the testing side (Contact has an initializer that takes a UIImage *).

tests/Contacts/ContactSortingTest.m
```diff
@@ -0,0 +1,119 @@
...
+    for (NSArray<NSString *>*namePair in namePairs) {
+        Contact *c = [[Contact alloc] initWithContactWithFirstName:namePair[0]
+                                                       andLastName:namePair[1]
+                                           andUserTextPhoneNumbers:@[]
+                                                          andImage:nil
+                                                      andContactID:fakeRecordId++];
+        [contacts addObject:c];
+    }
+    
+    return [NSArray arrayWithArray:contacts]; // Return an immutable for good hygene
```
michaelkirk: `[contacts copy]` is a more idiomatic way to return an immutable object from it's mutable variant.

michaelkirk (commented):
Tiny nit in the test case, but otherwise they read really clearly. Nice approach!

src/Contacts/Contact.h
```diff
@@ -1,5 +1,6 @@
 #import <AddressBook/AddressBook.h>
 #import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
```
michaelkirk: Can we get away with just forward declaring 'UIImage' instead?

tests/Contacts/ContactSortingTest.m
roostr: Sounds good, will force push a new branch with both instances changed once you complete the review.

src/Contacts/Contact.h
roostr: Yup, good idea. Will push that once you've completed the review.

roostr:
Pushed 0e530d0, which only differs from the previous by changing the new `[NSArray arrayWithArray]` to `array copy` and forward declares `UIImage` so a `UIKit` import is not necessary.

PR #68, opened 2016-11-26
OWS Contact from CNContact

michaelkirk:
* specify array types
* populate emails
* nullability annotations

// FREEBIE


michaelkirk:
This might stomp on #67...

roostr:
Cool, won't be hard to fix up #67

Issue #72, opened 2016-12-07
Steps towards share app extensions


guillett:
A currently missing feature of Signal on iOS is the ability to easily share images, links, etc. (cf. https://github.com/WhisperSystems/Signal-iOS/issues/650 and https://github.com/WhisperSystems/Signal-iOS/issues/1419)

I am looking at this and investigating the best way to get that feature. I am sharing my thoughts and progress here to start a conversation and make sure that we can add that feature to Signal.

So far (and to my understanding), in order to have an App extension:

- [ ] The underlying SQLite database of Signal needs to be sharable between apps (namely the Signal app and the App extension (two different binaries so different sandboxes))
- [ ] As App extensions can't use ```[UIApplication sharedApplication]```, some refactoring is required in SignalServiceKit

I have done most of the work in my repo and would like to push amended/cleaned code to the official one.

## 1 - Moving underlying database

Note: Sharable locations are managed via App Groups that are owned by registered users/groups.

- [ ] Have a working set up with the database in a shared location
- [ ] Set the project so that it is easy for developers to keep on contributing
- [ ] Ensure a smooth transition from the current location to the shared one.

Notes:
* It is a bit painful to work with App Groups.
  * Having different entitlements for DEBUG and RELEASE would probably help
  * App group names should probably not be hard coded and referred to via the embedded entitlement files.

## 2 - Allowing SignalServiceKit to be built as part of an App Extension

Some features in SignalServiceKit are not relevant/builtable for an App Extension. Those could be disabled via a macro definition. That's the approach used by AFNetworking (cf. https://github.com/AFNetworking/AFNetworking/issues/2119). A refactoring of the SignalServiceKit.podspec (addition of subspecs) seems to do the trick.

Notes:
* I *naively* put sections of code in ```#ifndef SIGNAL_SK_APP_EXTENSIONS``` and that does the trick.
* Having a clean podspec was a bit painful. For reference, I used the same approach as YapDatabase (i.e. ```s.xcconfig = { 'OTHER_CFLAGS' => '-DSIGNAL_SK_APP_EXTENSIONS' } ```)

jooize:
Perhaps Wire's experience is of use.

https://medium.com/@wireapp/the-challenge-of-implementing-ios-share-extension-for-end-to-end-encrypted-messenger-dd33b52b1e97

michaelkirk:
Per WhisperSystems/Signal-iOS#2341 we've moved
[...]

PR #73, opened 2016-12-08
Allow signal database to be in a shared folder

guillett:
This is a first formal attempt to allow signal database to be in a shared folder.

A shared location exists if the application is part of a group with the suffix "signal_service".
If there is no such group, the database is placed in a standalone location (current location).

I added some logic to allow App Group retrieval from different entitlement files to facilitate contributor work.

Note: App group names will have to be updated to group names *owned* by the proper development team (OWS).

I am more than happy to refactor/amend that code.


guillett:
@michaelkirk Did you have by a chance to have a look at #73 and/or #72? I am quite interested in your (OWS's) point of view regarding the proper steps towards the requested share feature.

michaelkirk:
Hey @guillet, Thanks for going ahead with this. I did see your pull requests. It’s a large change that deserves a lot of scrutiny. Realistically I won’t have time to dig into it for the next couple weeks, but I will eventually. 

I think it’s a really important feature and want to make sure we get it right. Thanks for getting the ball rolling.

michaelkirk:
Did you see this https://github.com/yapstudios/YapDatabase/wiki/Multiprocess-Support @guillett?

Seems like something that would need to be addressed.

guillett:
@michaelkirk You are totally right and I haven't had a look at that before.

To ensure that constraints are **consistently** satisfied it is probably better to have the whole correct set up embedded in SignalServiceKit (with no special setting in consumer projects [namely, Signal and App Extensions]).

guillett:
I am cleaning open PRs. As the project moved I close that one.

PR #74, opened 2016-12-15
check for errors in the keychain password retrieval

svevang:
In the course of running signal in a simulator, I ran into a situation where the database password stored in the keychain would not open my existing database.  This commit tests if accessing the keychain fails when fetching the db password and potentially prevents overwriting that password with a new one.

### DIFF
--- a/src/Storage/TSStorageManager.m
+++ b/src/Storage/TSStorageManager.m
@@ -238,22 +238,38 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 - (NSData *)databasePassword
 {
     [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
-    NSString *dbPassword = [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount];
 
-    if (!dbPassword) {
-        dbPassword = [[Randomness generateRandomBytes:30] base64EncodedString];
-        NSError *error;
-        [SAMKeychain setPassword:dbPassword forService:keychainService account:keychainDBPassAccount error:&error];
-        if (error) {
-            // Sync log to ensure it logs before exiting
-            NSLog(@"Exiting because we failed to set new DB password. error: %@", error);
-            exit(1);
-        } else {
-            DDLogError(@"Succesfully set new DB password. First launch?");
+    NSError *keyFetchError;
+    NSString *dbPassword =
+        [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
+
+    if (keyFetchError) {
+        switch (keyFetchError.code) {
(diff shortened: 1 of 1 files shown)


src/Storage/TSStorageManager.m
```diff
@@ -238,16 +238,20 @@ - (BOOL)databasePasswordAccessible
 - (NSData *)databasePassword
 {
     [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
-    NSString *dbPassword = [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount];
+
+    NSError *keyFetchError;
+    NSString *dbPassword = [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
+
+    if(keyFetchError) {
```
michaelkirk: spacing nit. `if (foo) {`

Or better yet, run git-clang-format.

svevang:
Thanks, I ran the change through `git clang-format` and squashed that into my commit.

michaelkirk:
This seems to crash new installs

> 2016-12-15 16:50:35.809 Signal[6936:52271] *** Terminating app due to uncaught exception 'Error when getting DB password.', reason: 'error: Error Domain=com.samsoffes.samkeychain Code=-25300 "The specified item could not be found in the keychain" UserInfo={NSLocalizedDescription=The specified item could not be found in the keychain}'

svevang:
Thanks! I just squashed a new commit and rebased.

Just to reiterate, the goal here is to prevent the accidental overwriting of the user's database password. In the case that a user's DB password _has_ changed, they would be stuck in a crash cycle (signal-ios boots, attempts to load db, bad password -> crash). The least-bad option in that case would be to blow away the inaccessible database and have the user register their account.

src/Storage/TSStorageManager.m
```diff
@@ -238,24 +238,40 @@ - (BOOL)databasePasswordAccessible
...
+    NSString *dbPassword =
+        [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
+
+    if (keyFetchError) {
+        switch (keyFetchError.code) {
+            case errSecItemNotFound:
+                dbPassword = [self createAndSetNewDatabasePassword];
+                break;
+            default:
+                [NSException raise:@"Serious error when getting DB password from keychain."
```
michaelkirk, line 252: Can you us a constant for the exception name, like `NSString *const TSStorageManagerInaccessiblePaswordException = @"TSStorageManagerInaccessiblePaswordException"`

and similarly in `createAndSetNewDatabasePassword`. That paves the way for any future handling based on name.

michaelkirk:
Thanks! Will be in 2.6.9.

PR #80, opened 2016-12-22
Added Omani country code

alimakki:
Added +968


michaelkirk:
This plus targeting regional domains included here: https://github.com/WhisperSystems/SignalServiceKit/pull/82

PR #82, opened 2017-01-03
Charlesmchen/censorship circumvention 2

charlesmchen:
This PR includes:

* Add censorship circumvention in Cuba, Oman per #1576.
* Add censorship circumvention in Iran per https://github.com/WhisperSystems/Signal-Android/issues/5992
* Update fronting to use a country-specific Google domain.
* I added Asserts.h.  I'd like to add it to the .pchs for SignalServiceKit and Signal-iOS.  
   * I'm wondering if we should modify SignalServiceKit to include an .xcodeproj file.

This PR does not include:

* This work isn't complete.  I'm not 100% clear on all of the changes that will be necessary.  
* I haven't tested - I'm not clear on how we'll test this.
* Do we need to bundle additional .crts with the app?  I suspect that GIAG2.crt is sufficient for all Google domains.

PTAL @michaelkirk

### DIFF
--- a/src/Network/OWSCensorshipConfiguration.m
+++ b/src/Network/OWSCensorshipConfiguration.m
@@ -6,14 +6,29 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NSString *const OWSCensorshipConfigurationFrontingHost = @"https://google.com";
 NSString *const OWSCensorshipConfigurationReflectorHost = @"signal-reflector-meek.appspot.com";
 
 @implementation OWSCensorshipConfiguration
 
-- (NSString *)frontingHost
+- (NSString *)frontingHost:(NSString *)e164PhonNumber
 {
-    return OWSCensorshipConfigurationFrontingHost;
+    OWSAssert(e164PhonNumber.length > 0);
+    
+    NSString *domain = nil;
+    for (NSString *countryCode in self.censoredCountryCodes.allKeys) {
+        if ([e164PhonNumber hasPrefix:countryCode]) {
+            domain = self.censoredCountryCodes[countryCode];
+        }
+    }
+    
+    // Fronting should only be used for countries specified in censoredCountryCodes,
+    // all of which have a domain specified.
+    OWSAssert(domain);
+    if (!domain) {
+        domain = @"google.com";
+    }
+    
+    return [@"https://" stringByAppendingString:domain];
 }
 
 - (NSString *)reflectorHost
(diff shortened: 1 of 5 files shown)


src/Network/OWSCensorshipConfiguration.m
```diff
@@ -6,33 +6,61 @@
...
+- (NSDictionary<NSString *, NSString *> *)censoredCountryCodes
 {
-    // Reports of censorship in:
-    // Egypt
-    // UAE
-    return @[@"+20",
-             @"+971"];
+    // Domain fronting should be used for the following countries.
+    //
+    // For each country, we should the appropriate google domain,
```
michaelkirk: I think you a word.

src/Util/Asserts.h
```diff
@@ -0,0 +1,30 @@
+//
+//  Asserts.h
+//
+//  Copyright (c) 2016 Open Whisper Systems. All rights reserved.
+//
+
```
michaelkirk, line 6: ci failure:
❌  /Users/travis/build/WhisperSystems/SignalServiceKit/src/Network/OWSCensorshipConfiguration.m:15:5: implicit declaration of function 'OWSAssert' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    OWSAssert(e164PhonNumber.length > 0);

Note that tests run in the context of `SignalServiceKit/Example/TSKitiOSTestApp/`

michaelkirk (changes requested):
Bulk of the changes look good!

src/Network/OWSCensorshipConfiguration.m
charlesmchen: Good catch, thanks.

src/Util/Asserts.h
charlesmchen: Sorry, this is what I meant about the .pch above.  I'm a fan of small, lean .pchs but this kind of asserts header is imo a strong candidate for inclusion, since ideally I'd like to use asserts in nearly every source file.

Anyhow, it sounds like that's out of the scope of this PR.  For now, I'll just fix by manually importing the header.

michaelkirk: > I'm a fan of small, lean .pchs but this kind of asserts header is imo a strong candidate for inclusion, since ideally I'd like to use asserts in nearly every source file.

blergh. Not your fault, missed that line. Yes, this makes sense for the .pch.

michaelkirk:
I'll leave it up to you as to if you wan't to move asserts into the pch, otherwise LGTM. :+1:

PR #83, opened 2017-01-04
Add asserts to .pch.

charlesmchen:
PTAL @michaelkirk 

Note: You'll need to do a `pod install` after this merge.


michaelkirk:
LGTM

PR #84, opened 2017-01-06
Add stub for analytics.

charlesmchen:
Adds a stub class for analytics that we can start using today, even though it is currently a no-op.  This allows us to begin tracking the places we'd like to use analytics (e.g. to track high-severity errors) if/when we ever put an analytics solution in place.

I stubbed this out since we'll want to use this in the "YapDatabase initialization failure" work, whose consequences are total data loss for the user.

* I also added a bare bones precommit script that can be used to rewrite the headers of any changed files.

PTAL @michaelkirk

### DIFF
--- a/Utilities/precommit.py
+++ b/Utilities/precommit.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+import subprocess 
+import datetime
+import argparse
+import commands
+
+
+git_repo_path = os.path.abspath(subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).strip())
+
+
+def splitall(path):
+    allparts = []
+    while 1:
+        parts = os.path.split(path)
+        if parts[0] == path:  # sentinel for absolute paths
+            allparts.insert(0, parts[0])
+            break
+        elif parts[1] == path: # sentinel for relative paths
+            allparts.insert(0, parts[1])
+            break
+        else:
+            path = parts[0]
+            allparts.insert(0, parts[1])
+    return allparts
+    
+    
+def process(filepath):
+    with open(filepath, 'rt') as f:
+        text = f.read()
+    original_text = text
+    
+    lines = text.split('\n')
+    while lines and lines[0].startswith('//'):
+        lines = lines[1:]
+    text = '\n'.join(lines)
+    text = text.strip()
+
+    header = '''//
+//  Copyright (c) %s Open Whisper Systems. All rights reserved.
+//
+
+''' % ( 
+    datetime.datetime.now().year,
+    )
+    text = header + text + '\n'
+
+    if original_text == text:
+        return
+    
+    short_filepath = filepath[len(git_repo_path):]
+    if short_filepath.startswith(os.sep):
+       short_filepath = short_filepath[len(os.sep):] 
+    print 'Updating:', short_filepath
+    
+    with open(filepath, 'wt') as f:
+        f.write(text)
+
+
+def should_ignore_path(path):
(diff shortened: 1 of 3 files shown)


Utilities/precommit.py
```diff
@@ -0,0 +1,123 @@
...
+    original_text = text
+    
+    lines = text.split('\n')
+    while lines and lines[0].startswith('//'):
+        lines = lines[1:]
+    text = '\n'.join(lines)
+    text = text.strip()
+
+    header = '''//
+//  %s
```
michaelkirk: I know it's a common convention, but I've never understood why people insist on including the file name in the header. I exclude it, but I'm not blocking on this =)

Utilities/precommit.py
```diff
@@ -0,0 +1,123 @@
...
+        
+    return False
+    
+    
+def process_if_appropriate(filepath):
+    filename = os.path.basename(filepath)
+    if filename.startswith('.'):
+        return
+    file_ext = os.path.splitext(filename)[1]
+    if file_ext not in ('.h', '.hpp', '.cpp', '.m', '.mm', '.pch'):
```
michaelkirk: dare you add a ".swift" for future compatibility? 😱

Utilities/precommit.py
```diff
@@ -0,0 +1,123 @@
...
+    ]
+    for ignore_path in ignore_paths:
+        if path.startswith(ignore_path):
+            return True
+    for component in splitall(path):
+        if component.startswith('.'):
+            return True
+        if component.endswith('.framework'):
+            return True
+        if component in ('Pods', 'ThirdParty',):
```
michaelkirk: Assuming we want to use this script wholesale in Signal-iOS, makes sense to add `Carthage`

src/Util/OWSAnalytics.h
```diff
@@ -0,0 +1,70 @@
+//
+//  OWSAnalytics.h
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+typedef NS_ENUM(NSUInteger, OWSAnalyticsSeverity) {
```
michaelkirk, line 7: reminder to `git clang-format` after staging.

michaelkirk (changes requested):

Utilities/precommit.py
charlesmchen: Done.

Utilities/precommit.py
charlesmchen: Done.

Utilities/precommit.py
charlesmchen: Done.

src/Util/OWSAnalytics.h
charlesmchen: I'm a bit concerned because although I don't have a definitive understanding about this, I've seen some warnings on the internet _not_ to use clang-format as you have it (i.e. only on the portions of a file changed by you) for the reasons I raised yesterday - i.e. formatting can't be done locally.  

But I guess you've been using it this way for a while without problems, so maybe I should just try it.  

Are you using something like this?

https://github.com/andrewseidl/githook-clang-format

charlesmchen:
PTAL @michaelkirk

src/Util/OWSAnalytics.h
michaelkirk: I think I did a `brew install clang-format`

michaelkirk:
LGTM

src/Util/OWSAnalytics.h
charlesmchen: Done. I've applied that to the entire contents of this branch.

PR #85, opened 2017-01-06
Feature/database errors

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/issues/1578

### DIFF
--- a/src/Storage/TSStorageManager.h
+++ b/src/Storage/TSStorageManager.h
@@ -24,7 +24,7 @@ extern NSString *const TSUIDatabaseConnectionDidUpdateNotification;
 - (void)setupDatabase;
 - (void)deleteThreadsAndMessages;
 - (BOOL)databasePasswordAccessible;
-- (void)wipeSignalStorage;
+- (void)resetSignalStorage;
 
 - (YapDatabase *)database;
 - (YapDatabaseConnection *)newDatabaseConnection;
--- a/src/Storage/TSStorageManager.m
+++ b/src/Storage/TSStorageManager.m
@@ -1,8 +1,12 @@
-//  Created by Frederic Jacobs on 27/10/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  TSStorageManager.m
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSStorageManager.h"
 #import "NSData+Base64.h"
+#import "OWSAnalytics.h"
 #import "OWSDisappearingMessagesFinder.h"
 #import "OWSReadReceipt.h"
 #import "SignalRecipient.h"
@@ -32,18 +36,24 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 
 @end
 
+#pragma mark -
+
 // Some lingering TSRecipient records in the wild causing crashes.
 // This is a stop gap until a proper cleanup happens.
 @interface TSRecipient : NSObject <NSCoding>
 
 @end
 
+#pragma mark -
+
 @interface OWSUnknownObject : NSObject <NSCoding>
 
 @end
 
+#pragma mark -
+
 /**
- * A default object to returned when we can't deserialize an object from YapDB. This can prevent crashes when
+ * A default object to return when we can't deserialize an object from YapDB. This can prevent crashes when
  * old objects linger after their definition file is removed. The danger is that, the objects can lay in wait
  * until the next time a DB extension is added and we necessarily enumerate the entire DB.
  */
@@ -61,9 +71,13 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 
 @end
 
+#pragma mark -
+
 @interface OWSUnarchiverDelegate : NSObject <NSKeyedUnarchiverDelegate>
 
 @end
+
+#pragma mark -
 
 @implementation OWSUnarchiverDelegate
 
@@ -74,6 +88,8 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 }
 
 @end
+
+#pragma mark -
 
 @implementation TSStorageManager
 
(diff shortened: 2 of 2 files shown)


src/Storage/TSStorageManager.m
```diff
@@ -92,24 +110,52 @@ + (instancetype)sharedManager {
 - (instancetype)initDefault
 {
     self = [super init];
+    
+    if (![self tryToLoadDatabase]) {
+        // Failing to load the database is a catastrophic.
```
michaelkirk: you a word again.

src/Storage/TSStorageManager.m
```diff
@@ -248,42 +296,58 @@ - (NSData *)databasePassword
...
-            default:
-                DDLogError(@"%@ Getting DB password from keychain failed with error: %@", self.tag, keyFetchError);
-                [NSException raise:TSStorageManagerExceptionNameDatabasePasswordInaccessible
-                            format:@"Getting DB password from keychain failed with error: %@", keyFetchError];
-                break;
+        // Either this is a new install so there's no existing password to retrieve
+        // or the keychain has become corrupt.  Either way, we want to get back to a
+        // "known good state" and behave like a new install.
+        
+        BOOL shouldHavePassword = [[NSUserDefaults standardUserDefaults] objectForKey:TSStorageManagerHasDatabasePassword];
```
michaelkirk: I think using `fileExistsAtPath` would get you the same information and save a couple lines farther down. It might even be more conceptually to-the-point since that's the file that this password corresponds to.

not a blocking concern.

src/Storage/TSStorageManager.m
```diff
@@ -248,42 +296,58 @@ - (NSData *)databasePassword
...
+        if (shouldHavePassword) {
+            OWSAnalyticsCriticalWithParameters(@"Could not retrieve database password from keychain",
+                                            @{
+                                              @"ErrorCode": @(keyFetchError.code)
+                                              });
         }
+        
+        [self deletePasswordFromKeychain];
```
michaelkirk: Same as before:

We are missing the call to `[TSAttachmentStream deleteAttachments];` which deletes any on-disk attachments.  Instead of these lines, `wipeSignalStorage` is *almost* what we want but won't work as is.

src/Storage/TSStorageManager.m
```diff
@@ -92,24 +110,52 @@ + (instancetype)sharedManager {
...
+    if (![self tryToLoadDatabase]) {
+        // Failing to load the database is a catastrophic.
+        //
+        // The best we can try to do is to discard the current database
+        // and behave like a clean install.
+        
+        OWSAnalyticsCritical(@"Could not load database");
+    
+        // Try to reset app by deleting database.
+        [self deletePasswordFromKeychain];
```
michaelkirk: We are missing the call to `[TSAttachmentStream deleteAttachments];` which deletes any on-disk attachments.  Instead of these lines, `wipeSignalStorage` is *almost* what we want but won't work as is - though see below for my comment there.

src/Storage/TSStorageManager.m
```diff
@@ -378,20 +442,29 @@ - (void)deleteThreadsAndMessages {
...
+    self.database = nil;
+    
+    [self deletePasswordFromKeychain];
+    
+    [self deleteDatabaseFile];
+    
     [TSAttachmentStream deleteAttachments];
-
+    
     [[self init] setupDatabase];
```
michaelkirk: I don't even think this `setupDatabase` bit is necessary, so maybe you *could* use this in multiple places above if you just remove this line.

It's unnecessary, since (IIRC) the only place this is called, we immediately kill the app.  In which case the database would be setup on next launch. (Users access this functionality in Settings > Delete Account)

Though *should* we be killing the app is a different question. Probably not.

src/Storage/TSStorageManager.m
```diff
@@ -92,24 +110,52 @@ + (instancetype)sharedManager {
...
+    
+    return self;
+}
 
+- (BOOL)tryToLoadDatabase {
+    
+    // We determine the database password first, since a side effect of
+    // this can be deleting any existing database file (if we're recovering
+    // from a corrupt keychain).
+    NSData *databasePassword = [self databasePassword];
```
michaelkirk, line 139: You know. THIS change made me curious, and I found out we were calling the cipher block which queries the keychain **25 times** in quick succession on startup. =/

michaelkirk (changes requested):

src/Storage/TSStorageManager.m
charlesmchen: Done.

src/Storage/TSStorageManager.m
charlesmchen: Okay, I split `wipeSignalStorage` into two methods to DRY this up.

src/Storage/TSStorageManager.m
charlesmchen: Hmm, you're right.  I was trying to avoid filesystem access, but "less state = more surface area for bugs."

src/Storage/TSStorageManager.m
charlesmchen: Yup, see above.

charlesmchen: Although... as discussed offline, I think we should convert our "media/blob" cache to a LRU cache where contents get evacuated based on some maximum size on disk.  My understanding is that the app disk usage currently grows in an unbounded way, which is:

* Burdensome on/troublesome for users.
* A vector of serious "out of disk space" bugs.

Many users keep their phones very close to full at all times.

src/Storage/TSStorageManager.m
charlesmchen: Huh!  Also, why was this calling `init`?!?  I'm going to take your suggestion remove this line and undo the "splitting" I mentioned above.

charlesmchen:
PTAL @michaelkirk

src/Util/OWSAnalytics.h
```diff
@@ -0,0 +1,70 @@
+//
+//  OWSAnalytics.h
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
```
michaelkirk, line 4: wanna run precommit on all these?

src/Storage/TSStorageManager.h
```diff
@@ -24,7 +24,7 @@ extern NSString *const TSUIDatabaseConnectionDidUpdateNotification;
 - (void)setupDatabase;
 - (void)deleteThreadsAndMessages;
 - (BOOL)databasePasswordAccessible;
-- (void)wipeSignalStorage;
+- (void)resetSignalStorage;
```
michaelkirk, line 27: Just a heads up that this is used in Signal-iOS. Generally if there are breaking changes in ssk, try to open the corresponding Signal-iOS PR that can run it simultaneously.

src/Storage/TSStorageManager.m
```diff
@@ -456,16 +448,15 @@ - (void)deleteDatabaseFile {
     }
 }
 
-- (void)wipeSignalStorage {
+- (void)resetSignalStorage {
     self.database = nil;
+    _dbConnection = nil;
```
michaelkirk, line 450: I prefer property getters wherever possible (even internally). My rationale is that some assumptions become easier when only getters, setters, dealloc, and init touch it.

I could be persuaded to *always* use ivars internally, but I don't think we should mix and match arbitrarily.

michaelkirk (approved):
LGTM

src/Storage/TSStorageManager.h
charlesmchen: Done https://github.com/WhisperSystems/Signal-iOS/pull/1594

src/Storage/TSStorageManager.m
charlesmchen: Sure, that sounds quite reasonable.  In fact, the only reason I accessed the ivar here is because this property is declared as `readonly` in the header and there's no `readwrite` property as one might expect in a private anonymous category.  Going forward I'll fix these as I find them!

PR #86, opened 2017-01-06
Add class to track app version.

charlesmchen:
Add an entity to track and report on app version history.  I've always found this kind of thing invaluable when debugging, to help notice that certain crashes only happen on clean installs, or certain issues only occur when upgrading from version X to Y, etc.

PTAL @michaelkirk

### DIFF
--- a/src/Util/AppVersion.m
+++ b/src/Util/AppVersion.m
@@ -1,0 +1,64 @@
+//
+//  AppVersion.m
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "AppVersion.h"
+
+@interface AppVersion ()
+
+@property (nonatomic) NSString *firstAppVersion;
+@property (nonatomic) NSString *lastAppVersion;
+@property (nonatomic) NSString *currentAppVersion;
+
+@end
+
+#pragma mark -
+
+@implementation AppVersion
+
++ (instancetype)instance {
+    static AppVersion *instance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        instance = [AppVersion new];
+        [instance configure];
+    });
+    return instance;
+}
+
+- (void)configure {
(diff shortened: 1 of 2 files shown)


src/Util/AppVersion.m
```diff
@@ -0,0 +1,64 @@
...
+    static AppVersion *instance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        instance = [AppVersion new];
+        [instance configure];
+    });
+    return instance;
+}
+
+- (void)configure {
```
michaelkirk, line 31: Did you see the similar logic at https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/environment/PropertyListPreferences.h#L53

Though we aren't currently tracking FirstAppVersion (it's a great addition).

This drives the "old" migration process here:
https://github.com/WhisperSystems/Signal-iOS/blob/master/Signal/src/environment/VersionMigrations.m#L36

Though since 2.6.4 I've moved away from that to individually runnable migrations, as I found them more wieldy:
https://github.com/WhisperSystems/Signal-iOS/tree/master/Signal/src/environment/Migrations

My plan is to some day pull the plug on people who haven't migrated and remove the old migration class, but as of yet the doors still open for the ~1% of users running pre 2.6.4.

I guess I bring this up because maybe this *could* work there, but you'd have to test it. The alternative is to just leave that redundant code untouched just to drive the migrations until we're comfortable marooning the rear guard.

michaelkirk (approved):
LGTM. :1st_place_medal:

src/Util/AppVersion.m
charlesmchen: Yes, at the risk of some redundancy, I wanted to include something that:

a) Was lightweight and could easily be made a dependency by any component in the system.
b) Is intended to be activated _very early_ in the app startup process.
c) Had certain specific side effects like logging.

I didn't want to touch something as important as migration logic, and since this logic is basically trivial, I wasn't too worried about the duplication.

PR #87, opened 2017-01-06
Remove Iran from censorship circumvention. Current approach isn't suf…

charlesmchen:
…ficient.

Per @moxie0 .

PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #88, opened 2017-01-11
Strip trailing whitespace in precommit script.

charlesmchen:
Not sure if you'll approve of this, since will clean up entire files, not just areas already touched. But we have a lot of trailing space in our files.  It seems to arise from a difference between XCode's auto format and clang-format's style, and clang-format doesn't seem to clean this. Thoughts?

PTAL @michaelkirk


michaelkirk:
I'm more worried about getting bad information from `git blame` than I am about having trailing white space. If it were possible to do it on the already changed lines I'd be all for it. I'm a :-1: on this as is.

PR #89, opened 2017-01-11
Add WebRTC setting.

charlesmchen:
Adds a preference that enables WebRTC-based calling.  In terms of SSK, this involves adding a property to the "update attributes". During the beta period, when this preference can be changed by the user, we'll need to plumb it through our SSK request logic.  That's this PR.  Later, we'l be able to revert much of this work and use a hard-coded value.

PTAL @michaelkirk

### DIFF
--- a/src/Account/TSAttributes.m
+++ b/src/Account/TSAttributes.m
@@ -13,18 +13,21 @@
 
 @implementation TSAttributes
 
-+ (NSDictionary *)attributesFromStorageWithVoiceSupport {
++ (NSDictionary *)attributesFromStorage:(BOOL)isWebRTCEnabled {
     return [self attributesWithSignalingKey:[TSStorageManager signalingKey]
-                            serverAuthToken:[TSStorageManager serverAuthToken]];
+                            serverAuthToken:[TSStorageManager serverAuthToken]
+                            isWebRTCEnabled:isWebRTCEnabled];
 }
 
 + (NSDictionary *)attributesWithSignalingKey:(NSString *)signalingKey
                              serverAuthToken:(NSString *)authToken
+                             isWebRTCEnabled:(BOOL)isWebRTCEnabled
 {
     return @{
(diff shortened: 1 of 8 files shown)


charlesmchen:
See https://github.com/WhisperSystems/Signal-iOS/pull/1597

src/Account/TSAttributes.m
```diff
@@ -13,18 +13,21 @@
...
 
 + (NSDictionary *)attributesWithSignalingKey:(NSString *)signalingKey
                              serverAuthToken:(NSString *)authToken
+                             isWebRTCEnabled:(BOOL)isWebRTCEnabled
 {
     return @{
         @"signalingKey" : signalingKey,
         @"AuthKey" : authToken,
-        @"voice" : [NSNumber numberWithBool:YES], // all Signal-iOS clients support voice
+        @"voice" : @(YES), // all Signal-iOS clients support voice
```
michaelkirk, line 29: Not being sure how a boolean object get's serialized into the ultimate HTTP request (e.g. `false` or `0` ?), I would be cautious here. 

Signal-iOS doesn't use this value when pulled from the contact intersection request anyway, but Signal-Android currently does (though apparently that's going away with the webrtc branch). So the failure case would be that these Signal-iOS users wouldn't be callable from Signal-Android.

Can you confirm it's the same? (or revert the change)

michaelkirk (commented):
Is there corresponding Signal-iOS code?

 WhisperSystems/Signal-iOS#1597 shows
screen shot 2017-01-12 at 1 11 35 pm

src/Account/TSAttributes.m
michaelkirk: ohhh. I guess I just didn't know what @(YES) does. Looks like it returns `[NSNumber numberWithBool:YES]` 😅

michaelkirk: LGTM

PR #91, opened 2017-01-12
Update SignalRecipient with “is WebRTC enabled” property from service.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/Threads/TSContactThread.m
+++ b/src/Contacts/Threads/TSContactThread.m
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 16/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSContactThread.h"
 #import "ContactsManagerProtocol.h"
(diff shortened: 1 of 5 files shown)


src/Contacts/Threads/TSContactThread.m
```diff
@@ -32,7 +32,12 @@ + (instancetype)getOrCreateThreadWithContactId:(NSString *)contactId
         [SignalRecipient recipientWithTextSecureIdentifier:contactId withTransaction:transaction];
 
     if (!recipient) {
-        recipient = [[SignalRecipient alloc] initWithTextSecureIdentifier:contactId relay:relay supportsVoice:YES];
+        recipient =
+            [[SignalRecipient alloc] initWithTextSecureIdentifier:contactId
+                                                            relay:relay
+                                                    supportsVoice:YES
+                                                   // Default to NO; ContactsUpdater will try to update this property.
```
michaelkirk, line 42: If the contactsUpdater does create a SignalRecipient with the correct `supportsWebRTC` value, aren't we clobbering it with a `NO` on line 48? ` [recipient saveWithTransaction:transaction];`

Maybe the save needs to be moved up. Or maybe it can be deleted, since it looks like ContactsUpdater saves the recipient.

michaelkirk (changes requested):

src/Contacts/Threads/TSContactThread.m
charlesmchen: Great catch. Fixed, and added a comment.  

Also, per our offline conversation, let's keep thinking about how we can:

* Avoid using concurrency unless necessary.
* Avoid concurrency issues when mutating our models.

charlesmchen:
PTAL @michaelkirk

src/Contacts/Threads/TSContactThread.m
michaelkirk: Agree. 

Some relevant reading that might shape our solutions https://github.com/YapStudios/YapDatabase/wiki#intermediate

michaelkirk:
LGTM 🌯

PR #94, opened 2017-01-17
Fix web socket issue.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Network/WebSockets/TSSocketManager.m
+++ b/src/Network/WebSockets/TSSocketManager.m
@@ -1,9 +1,5 @@
 //
-//  TSSocketManager.m
-//  TextSecureiOS
-//
-//  Created by Frederic Jacobs on 17/05/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "SubProtocol.pb.h"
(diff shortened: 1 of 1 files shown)


src/Network/WebSockets/TSSocketManager.m
```diff
@@ -30,11 +26,11 @@ @interface TSSocketManager ()
 
 @property (nonatomic, readonly, strong) OWSSignalService *signalService;
 
-@property (nonatomic, retain) NSTimer *pingTimer;
-@property (nonatomic, retain) NSTimer *reconnectTimer;
```
charlesmchen, line 34: `retain` is the default behavior for objects, so this is redundant.  99% of the time I use `@property (nonatomic)`. The exceptions are when I make a property `weak` or `readonly`.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -59,7 +55,6 @@ - (instancetype)init
     }
 
     _signalService = [OWSSignalService new];
-    _websocket = nil;
```
charlesmchen, line 62: Class properties default to nil, so this is redundant.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -110,9 +105,9 @@ - (void)becomeActive
     }
 
     // Discard the old socket which is already closed or is closing.
-    [socket close];
-    self.status = kSocketStatusClosed;
-    socket.delegate = nil;
+    [self closeWebSocket];
+
+    self.status = kSocketStatusConnecting;
```
charlesmchen, line 110: This was a bug (but not a regression) - we weren't immediately updating status when we created a new websocket.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -127,8 +122,15 @@ - (void)becomeActive
 }
 
 + (void)resignActivity {
-    SRWebSocket *socket = [[self sharedManager] websocket];
-    [socket close];
+    [[self sharedManager] closeWebSocket];
```
charlesmchen, line 125: This is the core fix - `resignActivity` was closing the websocket without nil'ing out the property, nil'ing out its delegate or updating the status property.

charlesmchen:
This PR is superceded by #95

PR #95, opened 2017-01-17
Feature/fix websocket2

charlesmchen:
Comments inline

PTAL @michaelkirk

### DIFF
--- a/src/Network/WebSockets/TSSocketManager.m
+++ b/src/Network/WebSockets/TSSocketManager.m
@@ -1,9 +1,5 @@
 //
-//  TSSocketManager.m
-//  TextSecureiOS
-//
-//  Created by Frederic Jacobs on 17/05/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "SubProtocol.pb.h"
@@ -26,27 +22,30 @@ NSString *const SocketOpenedNotification     = @"SocketOpenedNotification";
 NSString *const SocketClosedNotification     = @"SocketClosedNotification";
 NSString *const SocketConnectingNotification = @"SocketConnectingNotification";
 
+// TSSocketManager's properties should only be accessed from the main thread.
 @interface TSSocketManager ()
 
-@property (nonatomic, readonly, strong) OWSSignalService *signalService;
+@property (nonatomic, readonly) OWSSignalService *signalService;
 
-@property (nonatomic, retain) NSTimer *pingTimer;
-@property (nonatomic, retain) NSTimer *reconnectTimer;
+@property (nonatomic) NSTimer *pingTimer;
+@property (nonatomic) NSTimer *reconnectTimer;
 
 
-@property (nonatomic, retain) SRWebSocket *websocket;
+@property (nonatomic) SRWebSocket *websocket;
 @property (nonatomic) SocketStatus status;
 
 @property (nonatomic) UIBackgroundTaskIdentifier fetchingTaskIdentifier;
 
-@property BOOL didConnectBg;
-@property BOOL didRetreiveMessageBg;
-@property BOOL shouldDownloadMessage;
+@property (nonatomic) BOOL didConnectBg;
+@property (nonatomic) BOOL didRetreiveMessageBg;
+@property (nonatomic) BOOL shouldDownloadMessage;
 
-@property (nonatomic, retain) NSTimer *backgroundKeepAliveTimer;
-@property (nonatomic, retain) NSTimer *backgroundConnectTimer;
+@property (nonatomic) NSTimer *backgroundKeepAliveTimer;
+@property (nonatomic) NSTimer *backgroundConnectTimer;
 
 @end
+
+#pragma mark -
 
 @implementation TSSocketManager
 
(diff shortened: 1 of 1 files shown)


src/Network/WebSockets/TSSocketManager.m
```diff
@@ -26,28 +22,31 @@
 NSString *const SocketClosedNotification     = @"SocketClosedNotification";
 NSString *const SocketConnectingNotification = @"SocketConnectingNotification";
 
+// TSSocketManager's properties should only be accessed from the main thread.
 @interface TSSocketManager ()
 
-@property (nonatomic, readonly, strong) OWSSignalService *signalService;
```
charlesmchen: `strong` / `retain` is the default behavior for object properties, so this is redundant. 99% of the time I use `@property (nonatomic)`. The exceptions are when I make a property `weak` or `readonly`.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -57,9 +56,11 @@ - (instancetype)init
     if (!self) {
         return self;
     }
+    
+    OWSAssert([NSThread isMainThread]);
 
     _signalService = [OWSSignalService new];
-    _websocket = nil;
```
charlesmchen: Class properties default to nil, so this is redundant.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -87,6 +88,21 @@ + (void)becomeActive
 
 - (void)becomeActive
 {
+    OWSAssert([NSThread isMainThread]);
```
charlesmchen, line 91: I've moved all of the logic in this class that use class state to the main thread. Most everything was already there, so I don't think I've moved anything expensive.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -98,10 +114,12 @@ - (void)becomeActive
         switch ([socket readyState]) {
             case SR_OPEN:
                 DDLogVerbose(@"WebSocket already open on connection request");
+                OWSAssert(self.status == kSocketStatusOpen);
```
charlesmchen, line 117: I've added asserts around the key `status` property. The logic that updates the status here, for example, should be redundant but I've left it as it stands in order to be more redundant in production.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -110,9 +128,10 @@ - (void)becomeActive
     }
 
     // Discard the old socket which is already closed or is closing.
-    [socket close];
-    self.status = kSocketStatusClosed;
-    socket.delegate = nil;
+    [self closeWebSocket];
```
charlesmchen, line 131: A number of properties are closely related and should always be updated in unison. I've DRY'ed that up with closeWebSocket.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -110,9 +128,10 @@ - (void)becomeActive
...
+
+    OWSAssert(self.status == kSocketStatusClosed);
+    self.status = kSocketStatusConnecting;
```
charlesmchen, line 134: This was a notable omission.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -259,7 +305,13 @@ - (NSString *)webSocketAuthenticationString {
 }
 
 - (void)scheduleRetry {
-    if (![self.reconnectTimer isValid]) {
+    OWSAssert([NSThread isMainThread]);
+   
+    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
```
michaelkirk: Will this prevent us from reconnecting while running in the background due to receiving a remote notification?

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -127,13 +146,28 @@ - (void)becomeActive
 }
 
 + (void)resignActivity {
-    SRWebSocket *socket = [[self sharedManager] websocket];
-    [socket close];
+    OWSAssert([NSThread isMainThread]);
```
michaelkirk, line 149: This blows up in AppDelegate:

```
- (void)applicationWillResignActive:(UIApplication *)application {
       // ...Snip for brevity

    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
      if ([TSAccountManager isRegistered]) {
          dispatch_sync(dispatch_get_main_queue(), ^{
            [self protectScreen];
            [[[Environment getCurrent] signalsViewController] updateInboxCountLabel];
          });
          [TSSocketManager resignActivity]; // !!! <-- I suppose this could move up a line into the "main" block.
      }

       // ...Snip for brevity
    });
}
```

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -317,11 +383,17 @@ - (void)backgroundConnectTimerExpired {
 }
 
 - (void)backgroundTimeExpired {
-    [[self class] resignActivity];
+    OWSAssert([NSThread isMainThread]);
+    
+    if ([UIApplication sharedApplication].applicationState != UIApplicationStateActive) {
+        [[self class] resignActivity];
```
charlesmchen, line 416: This was a key bug. The app was often closing its websocket.

src/Network/WebSockets/TSSocketManager.m
michaelkirk: whoops. I see you've already addressed this in Signal-iOS:charlesmchen/fixWebsocket

:+1:

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Good catch, fixed.

michaelkirk:
LGTM, would you consider squashing these or adding some color to the individual commits? It's easier to get lost in the git history with identical commit messages.

charlesmchen:
> would you consider squashing these or adding some color to the individual commits

Done.

PR #96, opened 2017-01-18
Swallow errors returned by swift lint.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
sure. LGTM.

(There's no swift in SSK)

charlesmchen:
> (There's no swift in SSK)

Yes, but I use this script in other projects too like Signal-iOS

PR #98, opened 2017-01-22
Prevent session corruption by using same queue for encrypt vs. decrypt

michaelkirk:
// FREEBIE

**Corresponding PRs:**

Signal-iOS: https://github.com/WhisperSystems/Signal-iOS/pull/1630
SignalProtocol: https://github.com/WhisperSystems/SignalProtocolKit/pull/14

PTAL @charlesmchen

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -1,5 +1,6 @@
-//  Created by Michael Kirk on 10/7/16.
-//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "OWSMessageSender.h"
 #import "ContactsUpdater.h"
@@ -628,11 +629,17 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             // DEPRECATED - Remove after all clients have been upgraded.
             BOOL isLegacyMessage = ![message isKindOfClass:[OWSOutgoingSyncMessage class]];
 
-            NSDictionary *messageDict = [self encryptedMessageWithPlaintext:plainText
-                                                                toRecipient:recipient.uniqueId
-                                                                   deviceId:deviceNumber
-                                                              keyingStorage:[TSStorageManager sharedManager]
-                                                                     legacy:isLegacyMessage];
+            __block NSDictionary *messageDict;
+            // Mutating session state is not thread safe, so we operate on a serial queue, shared with decryption
+            // operations.
+            dispatch_sync([OWSDispatch sessionCipher], ^{
+                messageDict = [self encryptedMessageWithPlaintext:plainText
+                                                      toRecipient:recipient.uniqueId
+                                                         deviceId:deviceNumber
+                                                    keyingStorage:[TSStorageManager sharedManager]
+                                                           legacy:isLegacyMessage];
+            });
+
             if (messageDict) {
                 [messagesArray addObject:messageDict];
             } else {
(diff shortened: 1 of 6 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -628,11 +629,17 @@ - (void)sendSyncTranscriptForMessage:(TSOutgoingMessage *)message
...
 
-            NSDictionary *messageDict = [self encryptedMessageWithPlaintext:plainText
-                                                                toRecipient:recipient.uniqueId
-                                                                   deviceId:deviceNumber
-                                                              keyingStorage:[TSStorageManager sharedManager]
-                                                                     legacy:isLegacyMessage];
+            __block NSDictionary *messageDict;
+            // Mutating session state is not thread safe, so we operate on a serial queue, shared with decryption
+            // operations.
+            dispatch_sync([OWSDispatch sessionCipher], ^{
```
charlesmchen, line 635: Every time I see `dispatch_sync()`, I get worried about deadlock. Maybe we should start a CONCURRENCY.md which documents (for now) which threads/queues block on which threads/queues, so that we can do this kind of thing with confidence.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -218,17 +219,21 @@ - (void)processWebSocketRequestMessage:(WebSocketRequestMessage *)message {
...
+            if (!decryptedPayload) {
+                DDLogWarn(@"Failed to decrypt incoming payload or bad HMAC");
+                return;
+            }
 
-        OWSSignalServiceProtosEnvelope *envelope = [OWSSignalServiceProtosEnvelope parseFromData:decryptedPayload];
+            OWSSignalServiceProtosEnvelope *envelope = [OWSSignalServiceProtosEnvelope parseFromData:decryptedPayload];
 
-        [[TSMessagesManager sharedManager] handleReceivedEnvelope:envelope];
+            [[TSMessagesManager sharedManager] handleReceivedEnvelope:envelope];
```
charlesmchen: So it looks like `handleReceivedEnvelope` used to be invoked on the main thread but will now be invoked on the cipher queue.  Why not dispatch async to main thread?  That way:

* Only cipher work will be done on cipher queue.
* Any assumptions about concurrency in `handleReceivedEnvelope` won't be violated.
* Don't have to worry about `handleReceivedEnvelope` being expensive and blocking the cipher queue.

charlesmchen (commented):
Raised a couple of questions about concurrency.

src/Messages/OWSMessageSender.m
michaelkirk: My fear with documentation files is that the farther away they are from the code, the less likely they are to be kept up to date or read.

How do would you feel about my existing documentation here in the OWSDispatch class? https://github.com/WhisperSystems/SignalServiceKit/pull/98/files#diff-01540420599ea2ef7127ff1d06e00c81R17

src/Network/WebSockets/TSSocketManager.m
michaelkirk: Per your other comment, this was an obvious place to dispatch `async` (without further refactoring).

I could easily address what you've mentioned by getting ride of this `dispatch_async` call and  placing a couple more finely-grained `dispatch_sync` calls within this method.

src/Messages/OWSMessageSender.m
charlesmchen: Yeah, OWSDispatch might be a better place.  But it would just be nice to have a certain understanding:

* We should avoid dispatch_sync, only when necessary.
* We should have broad rules about which threads and queues can block on each other (i.e. dispatch sync).

src/Network/WebSockets/TSSocketManager.m
charlesmchen: I'd rather avoid dispatch_sync and do this:

```
dispatch_async(sessionCipher) {
// decryption logic
dispatch_async(main) {
[[TSMessagesManager sharedManager] handleReceivedEnvelope:envelope];
});
});
```

charlesmchen:
@michaelkirk  Is this going in the next release (2.6.13)?

michaelkirk:
PTAL @charlesmchen 

I left the one call to dispatch_sync in the message sender (encryption). I left this in because the existing error handling around this method is pretty gnarly. Making it async would only make it worse, or require a substantial refactor.

charlesmchen (approved):
LGTM

PR #99, opened 2017-01-23
dont reset storage before first unlock

michaelkirk:
Prevent destroying user database after restarting device, by not taking any drastic measures (like resetting the user's storage) when the app is running in the background.

If there's really a problem, it will manifest in the foreground, while in the background it's very likely the user just hasn't unlocked their device since restart.

corresponding Signal-iOS: https://github.com/WhisperSystems/Signal-iOS/pull/1636

PTAL @charlesmchen

### DIFF
--- a/src/Storage/TSStorageManager.h
+++ b/src/Storage/TSStorageManager.h
@@ -1,9 +1,5 @@
 //
-//  TSStorageManager.h
-//  TextSecureKit
-//
-//  Created by Frederic Jacobs on 27/10/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "TSStorageKeys.h"
@@ -21,9 +17,17 @@ extern NSString *const TSUIDatabaseConnectionDidUpdateNotification;
 @interface TSStorageManager : NSObject
 
 + (instancetype)sharedManager;
+
+/**
+ * Returns NO if:
+ *
+ * - Keychain is locked because device has just been restarted.
+ * - Password could not be retrieved because of a keychain error.
+ */
++ (BOOL)isDatabasePasswordAccessible;
+
 - (void)setupDatabase;
 - (void)deleteThreadsAndMessages;
-- (BOOL)databasePasswordAccessible;
 - (void)resetSignalStorage;
 
 - (YapDatabase *)database;
--- a/src/Storage/TSStorageManager.m
+++ b/src/Storage/TSStorageManager.m
@@ -1,6 +1,4 @@
 //
-//  TSStorageManager.m
-//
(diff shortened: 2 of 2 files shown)


src/Storage/TSStorageManager.h
```diff
@@ -21,9 +17,15 @@ extern NSString *const TSUIDatabaseConnectionDidUpdateNotification;
 @interface TSStorageManager : NSObject
 
 + (instancetype)sharedManager;
+
+/**
+ * Database password might not be accessible if the user just restarted the device and hasn't unlocked. If that's the
+ * case, we can't usefully start the app until they unlock.
```
charlesmchen: I think this comment could be a little misleading. Perhaps:

```
* Returns NO if:
*
* - Keychain is locked because device has just been restarted.
* - Password could not be retrieved because of a keychain error.
```

charlesmchen: The db password is created even before user registers, right? What if user de-registers? Is the db ever going to be deleted in that case?

src/Storage/TSStorageManager.m
```diff
@@ -287,6 +287,17 @@ - (BOOL)databasePasswordAccessible
...
 
+- (void)backgroundedAppDatabasePasswordInaccessibleWithError:(NSError *)error
+{
+    OWSAssert([UIApplication sharedApplication].applicationState == UIApplicationStateBackground);
+
+    // Presumably this happened in response to a push notification. It's possible that the keychain is corrupted
+    // but it could also just be that the user hasn't yet unlocked their device since our password is
+    // kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
+    [NSException raise:TSStorageManagerExceptionNameDatabasePasswordInaccessibleWhileBackgrounded
+                format:@"Unable to access database password. No unlock since device restart? Error: %@", error];
```
charlesmchen: Let's:

* Write error.description to the logs.
* Write [UIApplication sharedApplication].applicationState to the logs.
* Flush the logs.

src/Storage/TSStorageManager.m
```diff
@@ -296,7 +307,14 @@ - (NSData *)databasePassword
         [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
 
     if (keyFetchError) {
-        // Either this is a new install so there's no existing password to retrieve
```
charlesmchen, line 299: Wait, I take it back, before this second `if`, let's:

Let's:

* Write error.description to the logs.
* Write [UIApplication sharedApplication].applicationState to the logs.
* Flush the logs.

That way its in the logs whether or not we a) raise an exception or b) go to clean install mode.

src/Storage/TSStorageManager.m
charlesmchen: Strike that, this belongs below.

charlesmchen (approved):
LGTM

PR #100, opened 2017-01-25
Add assert macro that can be used in free functions.

charlesmchen:
Add "c" asserts that can be used in c-style free functions.

PTAL @michaelkirk


michaelkirk:
LGTM

PR #101, opened 2017-01-25
Disable resetting unreadable storage

michaelkirk:
Until we have better data on why this occurs, we don't want to risk
blowing users message archive.

// FREEBIE

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #102, opened 2017-01-27
Lookup methods yield SignalRecipient

michaelkirk:
Let's us remove some otherwise redundant code to look up the recipient
from their ID.

// FREEBIE

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #104, opened 2017-02-01
Rework concurrency in the signaling logic.

charlesmchen:
Adds an error, fixes a typo.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -436,7 +436,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
         }
 
         if ([exception.name isEqualToString:OWSMessageSenderRateLimitedException]) {
-            NSError *error = OWSErrorWithCodeDescription(OWSErrorCodeSingalServiceRateLimited,
+            NSError *error = OWSErrorWithCodeDescription(OWSErrorCodeSignalServiceRateLimited,
                 NSLocalizedString(@"FAILED_SENDING_BECAUSE_RATE_LIMIT",
                     @"action sheet header when re-sending message which failed because of too many attempts"));
             return failureHandler(error);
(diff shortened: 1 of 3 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -436,7 +436,7 @@ - (void)sendMessage:(TSOutgoingMessage *)message
         }
 
         if ([exception.name isEqualToString:OWSMessageSenderRateLimitedException]) {
-            NSError *error = OWSErrorWithCodeDescription(OWSErrorCodeSingalServiceRateLimited,
```
michaelkirk, line 439: gets me every singal time.

src/Util/OWSError.m
```diff
@@ -35,7 +37,13 @@
...
 {
     return OWSErrorWithCodeDescription(OWSErrorCodeFailedToSendOutgoingMessage,
-        NSLocalizedString(@"ERROR_DESCRIPTION_UNKNOWN_ERROR", @"Worst case generic error message"));
+                                       NSLocalizedString(@"ERROR_DESCRIPTION_UNKNOWN_ERROR", @"Worst case generic error message"));
+}
+
+NSError *OWSErrorMakeWebRTCMissingDataChannelError()
+{
+    return OWSErrorWithCodeDescription(OWSErrorCodeWebRTCMissingDataChannel,
+                                       NSLocalizedString(@"ERROR_DESCRIPTION_WEBRTC_MISSING_DATA_CHANNEL", @"Missing data channel while trying to sending message"));
```
michaelkirk: Is there any point in localizing an error like this? I don't think it would mean anything to our users or our translators.

michaelkirk (approved):
LGTM

src/Messages/OWSMessageSender.m
charlesmchen: =0

src/Util/OWSError.m
charlesmchen: Yeah, I was wondering the same thing.

PR #106, opened 2017-02-04
Automatically mark as read any messages sent by current user from ano…

charlesmchen:
…ther device.

Repro:

* Conduct a conversation between device A and desktop linked with device B.
* Read conversation on device B.

Expected: only "unread" messages from A are marked as unread.
Observed: messages sent from desktop linked with B show up as unread on B.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSMessagesManager.m
+++ b/src/Messages/TSMessagesManager.m
@@ -635,11 +635,10 @@ NS_ASSUME_NONNULL_BEGIN
 
           // Any messages sent from the current user - from this device or another - should be
           // automatically marked as read.
-          if ([TSAccountManager isRegistered]) {
-              BOOL shouldMarkMessageAsRead = [envelope.source isEqualToString:[TSAccountManager localNumber]];
-              if (shouldMarkMessageAsRead) {
-                  [incomingMessage markAsReadLocallyWithTransaction:transaction];
-              }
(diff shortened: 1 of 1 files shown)


src/Messages/TSMessagesManager.m
```diff
@@ -633,6 +633,15 @@ - (TSIncomingMessage *)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)
       if (thread && incomingMessage) {
           [incomingMessage saveWithTransaction:transaction];
 
+          // Any messages sent from the current user - from this device or another - should be
+          // automatically marked as read.
+          if ([TSAccountManager isRegistered]) {
```
michaelkirk: I think this `isRegistered` is just a sanity check, but it's written in a way that makes it seem like it's an expected case (if anything we should be exploding in the omitted `else` because something is very, very, wrong at this point). For example, we could never download the message we're processing if we weren't registered.

My concern is that it unnecessarily leads the reader down a crazy path, and we'd be best to omit it altogether.

michaelkirk (approved):
LGTM, good catch.

src/Messages/TSMessagesManager.m
charlesmchen: Good point. Changed it to be an assert.

PR #107, opened 2017-02-04
Order messages by "sent" timestamp, not "received" timestamp.

charlesmchen:
I'm not sure why, but we were ordering messages by "received" timestamp, not "sent" timestamp. This scrambles the ordering of conversations. 

Repro:

* Conduct a conversation between device A and desktop linked with device B.
* Read conversation on device B.  

Expected: Messages sent in order "alice, bob, carol, david"... will show up in same order.
Observed: Messages will be sent grouped by sender, since we fetch incoming and outgoing messages separately.

order-by-received
order-by-sent

PTAL @michaelkirk

### DIFF
--- a/src/Storage/TSDatabaseView.m
+++ b/src/Storage/TSDatabaseView.m
@@ -268,15 +268,6 @@ NSString *TSSecondaryDevicesDatabaseViewExtensionName = @"TSSecondaryDevicesData
 
 + (NSDate *)localTimeReceiveDateForInteraction:(TSInteraction *)interaction {
     NSDate *interactionDate = interaction.date;
-
-    if ([interaction isKindOfClass:[TSIncomingMessage class]]) {
-        TSIncomingMessage *message = (TSIncomingMessage *)interaction;
-
-        if (message.receivedAt) {
-            interactionDate = message.receivedAt;
-        }
-    }
-
     return interactionDate;
 }
 


michaelkirk:
fixes https://github.com/WhisperSystems/Signal-Desktop/issues/1043

michaelkirk:
A little git archeology shows us this was introduced here:

https://github.com/WhisperSystems/Signal-iOS/commit/7aad5c5971f9305fd4be7905e38b7d9fd10d12c6

> Fixing UX issue with unsynchronized clocks.
> 
>  TextSecure messages don’t get assigned a timestamp by the server. All
> of it is done end-to-end between both clients. A client could have a
> misconfigured clock or might want to forge a timestamp. Therefore, we
> address this issue by introducing a new receivedAt timestamp for
> incoming messages that will be used to sort the messages.

I think this concern is still relevant, though it's arguably less important than the issues it's causing since sync messages were introduced.  Another option would be to sort sync messages differently than we sort non-sync messages. 

Let's circle back with @moxie and see if Signal-Android has any conclusions.

charlesmchen:
Per offline conversation, a better fix might involve the service adding a "serviceReceivedAt:" timestamp to message envelops.

michaelkirk:
I think this can be closed as it's been superseded by #117.

PR #108, opened 2017-02-07
Missed calls increment unread counter/badge

michaelkirk:
This affects:
- per-thread badge in the inbox view
- app badge when viewed from springboard.
- back button unread count badge when viewing a different thread

// FREEBIE

PTAL @charlesmchen

### DIFF
--- a/src/Storage/TSDatabaseView.m
+++ b/src/Storage/TSDatabaseView.m
@@ -1,9 +1,5 @@
 //
-//  TSDatabaseView.m
-//  TextSecureKit
-//
-//  Created by Frederic Jacobs on 17/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "TSDatabaseView.h"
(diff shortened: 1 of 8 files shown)


charlesmchen:
Screenshots?

src/Storage/TSDatabaseView.m
```diff
@@ -35,17 +32,16 @@ + (BOOL)registerUnreadDatabaseView {
...
-              if (message.read == NO) {
-                  return message.uniqueThreadId;
-              }
-          }
-          return nil;
-        }];
+    YapDatabaseViewGrouping *viewGrouping = [YapDatabaseViewGrouping withObjectBlock:^NSString *(
+        YapDatabaseReadTransaction *transaction, NSString *collection, NSString *key, id object) {
+        if ([object conformsToProtocol:@protocol(OWSReadTracking)]) {
+            id<OWSReadTracking> possiblyRead = (id<OWSReadTracking>)object;
```
charlesmchen, line 38: This won't count "sync" messages, "read receipts", etc. will it?

charlesmchen:
I guess screenshots aren't necessary. Will the unread counter(c) get incremented when we receive a "security number changed" event?

src/Storage/TSDatabaseView.m
michaelkirk: Correct, we currently are not handling the eventual case where Signal-Desktop might know about call messages and then mark them as read.

charlesmchen (approved):
LGTM

PR #109, opened 2017-02-08
Log when we mark a message as read locally

michaelkirk:
This is in pursuit of some now (for me) un-reproducible behavior where
messages on some clients were being immediately marked as read.

// FREEBIE


charlesmchen (approved):
LGTM hope this fixes it.

michaelkirk:
Well, I'm not really expecting it to fix anything, but just to shed some light on what's happening.

PR #110, opened 2017-02-08
Improve the call status messages in conversation view.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/issues/1725

img_4292

Essentially, calls start in an "incomplete incoming" or "incomplete outgoing" state and are updated to "incoming" or "outgoing" when the call connects.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSCall.h
+++ b/src/Messages/TSCall.h
@@ -13,6 +13,9 @@ typedef enum {
     RPRecentCallTypeIncoming = 1,
     RPRecentCallTypeOutgoing,
     RPRecentCallTypeMissed,
+    // These call types are used until the call connects.
+    RPRecentCallTypeOutgoingIncomplete,
+    RPRecentCallTypeIncomingIncomplete,
 } RPRecentCallType;
 
 @interface TSCall : TSInteraction <OWSReadTracking>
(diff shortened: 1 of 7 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1735

src/Messages/TSCall.h
```diff
@@ -13,6 +13,9 @@ typedef enum {
     RPRecentCallTypeIncoming = 1,
     RPRecentCallTypeOutgoing,
     RPRecentCallTypeMissed,
+    // These call types are used until the call connects.
+    RPRecentCallTypeOutgoingIncomplete,
+    RPRecentCallTypeIncomingIncomplete,
```
charlesmchen, line 18: Incoming calls start as "incoming incomplete" and end up as "incoming" or "missed."
Outgoing calls start as "outgoing incomplete" and only become "outgoing" when they connect. If the outgoing call never connects (because you hang up first or the callee declines) the call simply remains in "outgoing incomplete".  It would be more consistent to rename:

* RPRecentCallTypeIncoming -> RPRecentCallTypeIncomingComplete
* RPRecentCallTypeOutgoing -> RPRecentCallTypeOutgoingComplete
* RPRecentCallTypeMissed -> RPRecentCallTypeIncomingFailed

And make a new state RPRecentCallTypeOutgoingFailed, but I didn't bother since this wouldn't have any visible effect.

michaelkirk (approved):
LGTM

PR #111, opened 2017-02-08
Avoid deadlock when marking self-sent messages as read.

michaelkirk:
Fetching the localNumber invokes another transaction, which can
result in deadlock if called during an existing transaction.

PTAL @charlesmchen 

I've been bitten by similar versions of this bug many times. There are broadly two approaches I see to avoiding it, both with their shortcomings...

1. Prefer methods that take a transaction as a parameter (this can lead to some really verbose code)
2. Make your transactions smaller (this can sometimes cause inconsistent state)


charlesmchen (commented):
LGTM thank you for catching and fixing this. 

One approach would be to label every single method in the codebase that uses a transaction directly or indirectly with a header comment, so that we could (eventually) verify the safety of every statement in a transaction.

PR #113, opened 2017-02-09
Mark "attempting out" messages as "unsent" on app launch.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/issues/1035

I tried to use `OWSDisappearingMessagesJob` and `OWSDisappearingMessagesFinder` as a model.  

* I used an index & query rather than a view.  Thoughts?
* I significantly elaborated the logging around message sending, failures and retries.
* Currently, if the device has no network access, message sending still waits through repeated retry attempts to timeout so messages don't fail for a couple of minutes. I wonder if we should abort retry in this case.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -409,6 +409,8 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             success:(void (^)())successHandler
             failure:(void (^)(NSError *error))failureHandler
 {
+    DDLogDebug(@"%@ sending message to service: %@", self.tag, message.debugDescription);
+
     if (remainingAttempts <= 0) {
         // We should always fail with a specific error.
         DDLogError(@"%@ Unexpected generic failure.", self.tag);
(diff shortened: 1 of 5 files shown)


charlesmchen:
See https://github.com/WhisperSystems/Signal-iOS/pull/1741

src/Messages/OWSMessageSender.m
```diff
@@ -554,6 +560,9 @@ - (void)handleMessageSentLocally:(TSOutgoingMessage *)message
 {
     [self saveMessage:message withState:TSOutgoingMessageStateSent];
     if (message.shouldSyncTranscript) {
+        // TODO: I suspect we shouldn't optimistically set hasSyncedTranscript.
+        //       We could set this in a success handler for [sendSyncTranscriptForMessage:].
```
charlesmchen, line 564: @michaelkirk thoughts?

charlesmchen:
PTAL @michaelkirk

src/Messages/OWSMessageSender.m
michaelkirk: I agree it's weird to optimistically set it like this. But make sure than any changes don't result in it being sent once per group recipient.

E.g. if I'm in a group with Alice, Bob, and Georgia, I should still send only one sync message.

michaelkirk (approved):
LGTM

src/Messages/OWSMessageSender.m
charlesmchen: I've opened a new issue where we can track/discuss this issue: https://github.com/WhisperSystems/Signal-iOS/issues/1761

PR #116, opened 2017-02-13
Add "hard reset git" script.

charlesmchen:
PTAL @michaelkirk


michaelkirk (approved):
in Signal-iOS this is called Scripts/HardResetGit.sh

LGTM

charlesmchen:
> in Signal-iOS this is called Scripts/HardResetGit.sh

Forgot I already made that. Closing this.

PR #117, opened 2017-02-15
Add “received at” timestamp to all TSMessages so that they may be sor…

charlesmchen:
A new version of https://github.com/WhisperSystems/SignalServiceKit/pull/107

We were only tracking "received at" timestamps for incoming messages.  I've modified the logic so that we now track this on all `TSMessages` using a "received at data" property.

* We could track this on just `TSIncomingMessages` and `TSOutgoingMessages`, but I chose to do it for all.  I think that makes sense... this will also affect the ordering of `TSErrorMessage`, `TSInfoMessage`, but my impression is that that will be beneficial or have no effect.

One oddity is that I don't want to lose the existing `receivedAt` values from `TSIncomingMessages`, so I've left that property and we use a `bestReceivedAtDate` method to try to use new timestamp but that falls back to the old timestamp if possible.

PTAL @michaelkirk

### DIFF
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -248,9 +248,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)updateWithLastMessage:(TSInteraction *)lastMessage transaction:(YapDatabaseReadWriteTransaction *)transaction {
     NSDate *lastMessageDate = lastMessage.date;
 
-    if ([lastMessage isKindOfClass:[TSIncomingMessage class]]) {
-        TSIncomingMessage *message = (TSIncomingMessage *)lastMessage;
-        lastMessageDate            = message.receivedAt;
+    if ([lastMessage isKindOfClass:[TSMessage class]]) {
+        TSMessage *message = (TSMessage *)lastMessage;
+        if ([message bestReceivedAtDate]) {
+            lastMessageDate = [message bestReceivedAtDate];
+        }
     }
 
     if (!_lastMessageDate || [lastMessageDate timeIntervalSinceDate:self.lastMessageDate] > 0) {
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -246,14 +246,7 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)updateWithLastMessage:(TSInteraction *)lastMessage transaction:(YapDatabaseReadWriteTransaction *)transaction {
-    NSDate *lastMessageDate = lastMessage.date;
-
-    if ([lastMessage isKindOfClass:[TSMessage class]]) {
-        TSMessage *message = (TSMessage *)lastMessage;
-        if ([message bestReceivedAtDate]) {
-            lastMessageDate = [message bestReceivedAtDate];
-        }
-    }
+    NSDate *lastMessageDate = [lastMessage receiptDateForSorting];
 
     if (!_lastMessageDate || [lastMessageDate timeIntervalSinceDate:self.lastMessageDate] > 0) {
         _lastMessageDate = lastMessageDate;
--- a/src/Messages/Interactions/TSIncomingMessage.m
+++ b/src/Messages/Interactions/TSIncomingMessage.m
@@ -61,7 +61,8 @@ NSString *const TSIncomingMessageWasReadOnThisDeviceNotification = @"TSIncomingM
 
     _authorId = authorId;
     _read = NO;
-    _receivedAt = [NSDate date];
+
+    OWSAssert(self.receivedAtDate);
 
     return self;
 }
@@ -132,6 +133,18 @@ NSString *const TSIncomingMessageWasReadOnThisDeviceNotification = @"TSIncomingM
     _read = YES;
     [self saveWithTransaction:transaction];
     [self touchThreadWithTransaction:transaction];
+}
+
+- (nullable NSDate *)bestReceivedAtDate
+{
+    NSDate *result = [super bestReceivedAtDate];
+    if (!result) {
+        // For backward compatibility with messages received before
+        // TSMessage.receivedAtData was added, honor TSIncomingMessage.receivedAt
+        // if TSMessage.receivedAtData is not set.
+        result = self.receivedAt;
+    }
+    return result;
 }
 
 #pragma mark - Logging
--- a/src/Messages/Interactions/TSIncomingMessage.m
+++ b/src/Messages/Interactions/TSIncomingMessage.m
@@ -135,18 +135,6 @@ NSString *const TSIncomingMessageWasReadOnThisDeviceNotification = @"TSIncomingM
     [self touchThreadWithTransaction:transaction];
 }
 
-- (nullable NSDate *)bestReceivedAtDate
-{
-    NSDate *result = [super bestReceivedAtDate];
-    if (!result) {
-        // For backward compatibility with messages received before
-        // TSMessage.receivedAtData was added, honor TSIncomingMessage.receivedAt
-        // if TSMessage.receivedAtData is not set.
-        result = self.receivedAt;
-    }
-    return result;
-}
-
 #pragma mark - Logging
 
 + (NSString *)tag
--- a/src/Messages/Interactions/TSMessage.h
+++ b/src/Messages/Interactions/TSMessage.h
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 12/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSInteraction.h"
 
@@ -29,6 +30,9 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 @property (nonatomic, readonly) uint64_t expiresAt;
 @property (nonatomic, readonly) BOOL isExpiringMessage;
 @property (nonatomic, readonly) BOOL shouldStartExpireTimer;
+// _DO NOT_ access this property directly.  You almost certainly
+// want to use bestReceivedAtDate instead.
+@property (nonatomic, readonly) NSDate *receivedAtDate;
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp;
 
@@ -59,6 +63,11 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 - (nullable instancetype)initWithCoder:(NSCoder *)coder NS_DESIGNATED_INITIALIZER;
 
 - (BOOL)hasAttachments;
+
+// This message should return TSMessage.receivedAtDate for most messages.
+// For messages received before TSMessage.receivedAtDate was added, this
+// will try to return TSIncomingMessage.receivedAt.
+- (nullable NSDate *)bestReceivedAtDate;
 
 @end
 
--- a/src/Messages/Interactions/TSMessage.h
+++ b/src/Messages/Interactions/TSMessage.h
@@ -31,7 +31,7 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 @property (nonatomic, readonly) BOOL isExpiringMessage;
 @property (nonatomic, readonly) BOOL shouldStartExpireTimer;
 // _DO NOT_ access this property directly.  You almost certainly
-// want to use bestReceivedAtDate instead.
+// want to use receiptDateForSorting instead.
 @property (nonatomic, readonly) NSDate *receivedAtDate;
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp;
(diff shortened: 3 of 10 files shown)


charlesmchen:
Fixes: https://github.com/WhisperSystems/Signal-iOS/issues/1772

src/Messages/Interactions/TSMessage.m
```diff
@@ -132,6 +133,10 @@ - (nullable instancetype)initWithCoder:(NSCoder *)coder
     }
 
     _schemaVersion = OWSMessageSchemaVersion;
+
+    // We _DO NOT_ set _receivedAt_ in this constructor.  We don't want to
+    // set the receivedAt time for old messages in the data store.
```
michaelkirk: Why not? More on this below in the YapView changes.

src/Storage/TSDatabaseView.m
```diff
@@ -265,11 +266,10 @@ + (void)asyncRegisterSecondaryDevicesDatabaseView
 + (NSDate *)localTimeReceiveDateForInteraction:(TSInteraction *)interaction {
     NSDate *interactionDate = interaction.date;
 
-    if ([interaction isKindOfClass:[TSIncomingMessage class]]) {
-        TSIncomingMessage *message = (TSIncomingMessage *)interaction;
-
-        if (message.receivedAt) {
-            interactionDate = message.receivedAt;
+    if ([interaction isKindOfClass:[TSMessage class]]) {
```
michaelkirk: How are we going to sort existing messages by bestReceivedAt if it's nil?

Two alternatives:
- The default implementation of `bestReceivedAtDate` could return `interaction.date` if `receivedAt` is nil.
OR
- Do the same thing, but in the NSCoder initializer

As is, it might work for now, but I suspect the next time we bump the version on this view we'll lose all sorting for the old messages.

Though you've done a good job of commenting the properties, I'd lean towards migrating the data, so that down the road we're less likely to be surprised by a nil field that "should never be nil".

src/Messages/Interactions/TSMessage.h
```diff
@@ -60,6 +64,11 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 
 - (BOOL)hasAttachments;
 
+// This message should return TSMessage.receivedAtDate for most messages.
+// For messages received before TSMessage.receivedAtDate was added, this
+// will try to return TSIncomingMessage.receivedAt.
+- (nullable NSDate *)bestReceivedAtDate;
```
michaelkirk: nit: I'm having a hard time keeping track of the 3 similar-but-different fields in the inheritance hierarchy:

2. TSIncomingMessage.receivedAt (local time when message was received)
3. TSMessage.receivedAtDate
4. TSMessage.bestReceivedAtDate

Since we don't really "Receive" outgoing messages (except synced transcripts), and since it's only "best" in the context of sorting, what do you think of calling it something like `sortingDate` to differentiate it from the others?

src/Messages/Interactions/TSMessage.m
```diff
@@ -102,6 +102,7 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
     _expiresInSeconds = expiresInSeconds;
     _expireStartedAt = expireStartedAt;
     [self updateExpiresAt];
+    _receivedAtDate = [NSDate date];
```
michaelkirk, line 105: I think this is the centerpiece of the change, right?

By tracking local clock time of creation, instead of envelope timestamp, we can order the creation of these interactions.

(unless the user changes their clock)

Does this do the right thing for sync transcript messages (e.g. messages that you sent via desktop) received late? I guess if the server takes care to deliver us the messages in the order it received them?

One nit:  can we call it `createdAtDate` since the existing name is so close to the similar but different `receivedAt`?

src/Messages/Interactions/TSIncomingMessage.m
```diff
@@ -134,6 +135,18 @@ - (void)markAsReadWithoutNotificationWithTransaction:(YapDatabaseReadWriteTransa
     [self touchThreadWithTransaction:transaction];
 }
 
+- (nullable NSDate *)bestReceivedAtDate
+{
+    NSDate *result = [super bestReceivedAtDate];
+    if (!result) {
+        // For backward compatibility with messages received before
+        // TSMessage.receivedAtData was added, honor TSIncomingMessage.receivedAt
```
michaelkirk: typo: TSMessage.receivedAtDat_a_

src/Contacts/TSThread.m
```diff
@@ -248,9 +248,11 @@ - (NSString *)lastMessageLabel {
 - (void)updateWithLastMessage:(TSInteraction *)lastMessage transaction:(YapDatabaseReadWriteTransaction *)transaction {
     NSDate *lastMessageDate = lastMessage.date;
 
-    if ([lastMessage isKindOfClass:[TSIncomingMessage class]]) {
```
michaelkirk: Just trying to understand, does this actually change any behavior?

michaelkirk:
Fixes https://github.com/WhisperSystems/Signal-iOS/issues/1398

src/Messages/Interactions/TSMessage.m
charlesmchen: > Does this do the right thing for sync transcript messages (e.g. messages that you sent via desktop) received late? I guess if the server takes care to deliver us the messages in the order it received them?

Yes, there's a single queue so this should work.  

> By tracking local clock time of creation, instead of envelope timestamp...

Right, we were already doing this for incoming messages and this just generalizes that to all `TSMessages`.

> One nit: can we call it createdAtDate since the existing name is so close to the similar but different receivedAt?

I'm happy to change the name, but I feel like `createdAtDate` is a bit misleading.  "Created" evokes (at least for me) when the message was created by the sender.

src/Messages/Interactions/TSMessage.m
charlesmchen: `receivedAt` is superceded by `receivedAtDate`.  I suppose we could continue to set it, but it would serve no value - and possibly create confusion.  The message would then have two timestamps that were always _almost_ the same but always off by a few ms.  We would never want to consult `receivedAt`, so why set it?

src/Messages/Interactions/TSIncomingMessage.m
charlesmchen: Done.

src/Contacts/TSThread.m
charlesmchen: Yes, I believe this logic is used to update the "last message date" for threads, which determines their ordering in the conversations view.  We've always updated that ordering for any interaction, but we were only using accurate "received" timestamps for incoming messages. Going forward we'll use "received" timestamps for all interactions. That should make the ordering of conversations more consistent.

michaelkirk: I don't think that these several lines are determining the order of anything - rather just what's displayed in the inbox view, which shows the last message for each thread.

src/Storage/TSDatabaseView.m
charlesmchen: > How are we going to sort existing messages by bestReceivedAt if it's nil?

1. We try to use `bestReceivedAt`. If its not set (because this is a message received before these changes)...
2. We try to use `bestReceived`.  Legacy messages only set that for incoming messages so if its not set...
3. We fallback to `date`.

Previously we did steps 2 & 3.  We're adding 1.

> The default implementation of `bestReceivedAtDate` could return `interaction.date` if `receivedAt` is nil.

I've taken your suggestions and a) DRY'd up the "default to TSInteraction.date" logic b) renamed `bestReceivedAtDate` to `receiptDateForSorting`. c) Pulled up `receiptDateForSorting` into `TSInteraction`.

> As is, it might work for now, but I suspect the next time we bump the version on this view we'll lose all sorting for the old messages.

I don't understand why?

src/Messages/Interactions/TSMessage.h
charlesmchen: Okay, I've removed the `receivedAt` property and added a migration. Please confirm that it is correct, since I haven't done a YapDatabase migration before.  

Also, `bestReceivedAtDate` (now called `receiptDateForSorting`) is a method, not a field.

charlesmchen:
PTAL @michaelkirk

src/Messages/Interactions/TSMessage.m
michaelkirk: Yeah, that makes sense. I think when I wrote that comment I was confusing `receivedAt` with `receivedAtDate`. Disregard...

src/Contacts/TSThread.m
charlesmchen: Huh, well seems to me that we should be using these dates for the conversation ordering as well.

src/Messages/Interactions/TSMessage.m
michaelkirk: I'm not that in love with `createdAt` you could pick something else, I'm just easily confused by the nuances of a field called `recordedAtDate` on an object which already has a field called `recordedAt` which is a date.

src/Messages/Interactions/TSMessage.m
```diff
@@ -132,6 +132,12 @@ - (nullable instancetype)initWithCoder:(NSCoder *)coder
         _attachmentIds = [NSMutableArray new];
     }
 
+    if (!_receivedAtDate) {
+        // TSIncomingMessage.receivedAt has been superceded by TSMessage.receivedAtDate.
```
michaelkirk, line 136: nit: If this is going to be a no-op for every class except TSIncomingMessage, maybe this belongs in TSIncomingMessage

src/Storage/TSDatabaseView.m
```diff
@@ -264,16 +264,7 @@ + (void)asyncRegisterSecondaryDevicesDatabaseView
...
-
-    if ([interaction isKindOfClass:[TSMessage class]]) {
-        TSMessage *message = (TSMessage *)interaction;
-        if ([message bestReceivedAtDate]) {
-            interactionDate = [message bestReceivedAtDate];
-        }
-    }
-
-    return interactionDate;
+    return [interaction receiptDateForSorting];
```
michaelkirk, line 267: Yay, I like this.

michaelkirk (approved):
LGTM

src/Messages/Interactions/TSMessage.m
michaelkirk: (the naming is just a nit btw) everything else LGTM

charlesmchen: > I'm not that in love with createdAt you could pick something else, I'm just easily confused by the nuances of a field called recordedAtDate on an object which already has a field called recordedAt which is a date.

I'm hoping that this isn't an issue any more, now that I've removed the `recordedAt` property?

src/Contacts/TSThread.m
michaelkirk: > Huh, well seems to me that we should be using these dates for the conversation ordering as well.

"these dates" **are** used for sorting, but that's not in this method. The sorting is in the YapDatabaseView.

src/Messages/Interactions/TSMessage.m
charlesmchen: I know you prefer `readonly` properties, which make it hard for TSIncomingMessage to set this property.  Can I leave this?

src/Storage/TSDatabaseView.m
michaelkirk: You're right, it was fine. 

For me anyway, the new implementation is clearer:
https://github.com/WhisperSystems/SignalServiceKit/pull/117/commits/99e85ecd1e48d953a7c3fff9c4afca20a9b1fada

src/Messages/Interactions/TSMessage.m
michaelkirk: Good point! I retract my nit.

src/Messages/Interactions/TSMessage.m
```diff
@@ -131,7 +132,17 @@ - (nullable instancetype)initWithCoder:(NSCoder *)coder
...
+        NSDate *receivedAt = [coder decodeObjectForKey:@"receivedAt"];
+        _receivedAtDate = receivedAt;
+    }
+
     _schemaVersion = OWSMessageSchemaVersion;
+
+    // We _DO NOT_ set _receivedAt_ in this constructor.  We don't want to
+    // set the receivedAt time for old messages in the data store.
```
michaelkirk: This comment can probably go away?

OK. seriously though, last nit from me. Thanks for your patience, I'm really happy with where this ended up.

src/Storage/TSDatabaseView.m
charlesmchen: > For me anyway, the new implementation is clearer:

There can be no doubt. I was as usual erring on the side of the more minimal change since I still feed new to this codebase, but doing it "the right way" was a much better idea.

src/Messages/Interactions/TSMessage.m
charlesmchen: Done. Thank you for the thorough review - it's a much better PR for your feedback.

PR #119, opened 2017-02-18
dedupe incoming messages

michaelkirk:
On Android when resending to a failed recipient, the entire group is broadcasted, which can result in every member of the group being spammed.

Here we dedupe incoming messages based on timestamp, sender Id, and the sender's device id.

Feature is in the first commit, the cleanup commit is unrelated, but something I noticed while working in the proximity.

PTAL @charlesmchen

### DIFF
--- a/src/Messages/Interactions/TSIncomingMessage.h
+++ b/src/Messages/Interactions/TSIncomingMessage.h
@@ -37,31 +37,6 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
                       messageBody:(nullable NSString *)body;
 
 /**
- *  Inits an incoming group message with attachments
- *
- *  @param timestamp
- *    When the message was created in milliseconds since epoch
- *  @param thread
- *    Thread to which the message belongs
- *  @param authorId
- *    Signal ID (i.e. e164) of the user who sent the message
- *  @param sourceDeviceId
- *    Numeric ID of the device used to send the message. Used to detect duplicate messages.
- *  @param body
- *    Body of the message
- *  @param attachmentIds
- *    The uniqueIds for the message's attachments
- *
- *  @return initiated incoming group message
- */
-- (instancetype)initWithTimestamp:(uint64_t)timestamp
-                         inThread:(TSThread *)thread
-                         authorId:(NSString *)authorId
-                   sourceDeviceId:(uint32_t)sourceDeviceId
-                      messageBody:(nullable NSString *)body
-                    attachmentIds:(NSArray<NSString *> *)attachmentIds;
-
-/**
  *  Inits an incoming group message that expires.
  *
  *  @param timestamp
@@ -93,8 +68,8 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
 
 
 /**
- * For sake of a smaller API, you must specify an author id for all incoming messages
- * though we technically could get the author id from a contact thread.
+ * For sake of a smaller API, and simplifying assumptions elsewhere, you must specify an author id for *all* incoming
+ * messages, even though we technically could infer the author id for a contact threads.
  */
 - (instancetype)initWithTimestamp:(uint64_t)timestamp NS_UNAVAILABLE;
 - (instancetype)initWithTimestamp:(uint64_t)timestamp inThread:(nullable TSThread *)thread NS_UNAVAILABLE;
--- a/src/Messages/Interactions/TSIncomingMessage.h
+++ b/src/Messages/Interactions/TSIncomingMessage.h
@@ -23,6 +23,8 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
  *    Thread to which the message belongs
  *  @param authorId
  *    Signal ID (i.e. e164) of the user who sent the message
+ *  @param sourceDeviceId
+ *    Numeric ID of the device used to send the message. Used to detect duplicate messages.
  *  @param body
  *    Body of the message
  *
@@ -31,6 +33,7 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(TSThread *)thread
                          authorId:(NSString *)authorId
+                   sourceDeviceId:(uint32_t)sourceDeviceId
                       messageBody:(nullable NSString *)body;
 
 /**
@@ -42,6 +45,8 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
  *    Thread to which the message belongs
  *  @param authorId
  *    Signal ID (i.e. e164) of the user who sent the message
+ *  @param sourceDeviceId
+ *    Numeric ID of the device used to send the message. Used to detect duplicate messages.
  *  @param body
  *    Body of the message
  *  @param attachmentIds
@@ -52,6 +57,7 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(TSThread *)thread
                          authorId:(NSString *)authorId
+                   sourceDeviceId:(uint32_t)sourceDeviceId
                       messageBody:(nullable NSString *)body
                     attachmentIds:(NSArray<NSString *> *)attachmentIds;
 
@@ -64,6 +70,8 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
  *    Thread to which the message belongs
  *  @param authorId
  *    Signal ID (i.e. e164) of the user who sent the message
+ *  @param sourceDeviceId
+ *    Numeric ID of the device used to send the message. Used to detect duplicate messages.
  *  @param body
  *    Body of the message
  *  @param attachmentIds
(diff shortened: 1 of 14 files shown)


src/Messages/Interactions/TSIncomingMessage.h
```diff
@@ -84,8 +68,8 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
 
 
 /**
- * For sake of a smaller API, you must specify an author id for all incoming messages
- * though we technically could get the author id from a contact thread.
+ * For sake of a smaller API, and simplifying assumptions elsewhere, you must specify an author id for *all* incoming
+ * messages, even though we technically could infer the author id for a contact threads.
```
charlesmchen: nit: extra preposition?
`...the author id for contact threads.`
or perhaps 
`...the author id of contact threads.`
Are contact threads 1:1 conversations?  Making sure I get the nomenclature.

src/Messages/Interactions/TSIncomingMessage.h
```diff
@@ -120,6 +104,7 @@ extern NSString *const TSIncomingMessageWasReadOnThisDeviceNotification;
 + (nullable instancetype)findMessageWithAuthorId:(NSString *)authorId timestamp:(uint64_t)timestamp;
 
 @property (nonatomic, readonly) NSString *authorId;
+@property (nonatomic, readonly) UInt32 sourceDeviceId;
```
charlesmchen: And it's okay that older messages in the data store will have a source device id of zero?

charlesmchen: I see, primary has device id 1, other devices are higher, so zero is "unknown".

src/Messages/TSMessagesManager.m
```diff
@@ -281,6 +284,15 @@ - (void)handlePreKeyBundleAsync:(OWSSignalServiceProtosEnvelope *)preKeyEnvelope
 - (void)handleEnvelope:(OWSSignalServiceProtosEnvelope *)envelope plaintextData:(NSData *)plaintextData
 {
     OWSAssert([NSThread isMainThread]);
+
+    BOOL duplicateEnvelope = [self.incomingMessageFinder existsMessageWithTimestamp:envelope.timestamp
+                                                                           sourceId:envelope.source
+                                                                     sourceDeviceId:envelope.sourceDevice];
+    if (duplicateEnvelope) {
```
charlesmchen, line 294: Should we constant-ize:
```
static uint32_t const OWSDeviceUnknownDeviceId = 0;
```  
and skip the duplicate check if:
```
if (envelope.sourceDevice != OWSDeviceUnknownDeviceId) {
   GPHAssert(envelope.sourceDevice >= OWSDevicePrimaryDeviceId);
}
```

src/Storage/OWSIncomingMessageFinder.m
```diff
@@ -0,0 +1,156 @@
...
+
+    YapDatabaseSecondaryIndexHandler *handler = [YapDatabaseSecondaryIndexHandler withObjectBlock:block];
+
+    return [[YapDatabaseSecondaryIndex alloc] initWithSetup:setup handler:handler];
+}
+
+- (void)asyncRegisterExtension
+{
+    DDLogInfo(@"%@ registering async.", self.tag);
+    [self.database registerExtension:self.indexExtension withName:OWSIncomingMessageFinderExtensionName];
```
charlesmchen: This method is called `asyncRegisterExtension` but it uses `asyncRegisterExtension`, not `registerExtension`.  Was it supposed to be async?  Can we combine with `registerExtension`?

src/Storage/OWSIncomingMessageFinder.m
```diff
@@ -0,0 +1,156 @@
...
+}
+
+- (void)registerExtension
+{
+    OWSAssert(NO);
```
charlesmchen: Why is this assert here?

src/Storage/OWSIncomingMessageFinder.m
```diff
@@ -0,0 +1,156 @@
...
+- (BOOL)existsMessageWithTimestamp:(uint64_t)timestamp
+                          sourceId:(NSString *)sourceId
+                    sourceDeviceId:(uint32_t)sourceDeviceId
+{
+    if (![self.database registeredExtension:OWSIncomingMessageFinderExtensionName]) {
+        DDLogError(@"%@ in %s but extension is not registered", self.tag, __PRETTY_FUNCTION__);
+        OWSAssert(NO);
+
+        // we should be initializing this at startup rather than have an unexpectedly slow lazy setup at random.
+        [self registerExtension];
```
charlesmchen, line 120: Is this to handle the race condition where we receive an incoming message before `asyncRegisterExtension` is called?  `asyncRegisterExtension` will always be called - is there any harm to registering twice?  Or is that why `asyncRegisterExtension` is actually synchronous?  If so, can't we remove this check? etc.

tests/Contacts/TSThreadTest.m
```diff
@@ -38,6 +40,7 @@ - (void)testDeletingThreadDeletesInteractions
     TSIncomingMessage *incomingMessage = [[TSIncomingMessage alloc] initWithTimestamp:10000
                                                                              inThread:thread
                                                                              authorId:@"fake-author-id"
+                                                                       sourceDeviceId:1
```
charlesmchen: Maybe use the constant?

tests/Storage/OWSOrphanedDataCleanerTest.m
```diff
@@ -45,6 +47,7 @@ - (void)testInteractionsWithoutThreadAreDeleted
     TSIncomingMessage *incomingMessage = [[TSIncomingMessage alloc] initWithTimestamp:1
                                                                              inThread:unsavedThread
                                                                              authorId:@"fake-author-id"
+                                                                       sourceDeviceId:1
```
charlesmchen: Ditto

src/Messages/Interactions/TSIncomingMessage.h
michaelkirk: Thanks! Fixed.

>Are contact threads 1:1 conversations? Making sure I get the nomenclature.

Yup, Thread, is effectively an abstract base class, implemented by either a ContactThread or a GroupThread.

src/Messages/Interactions/TSIncomingMessage.h
michaelkirk: Yup, should be fine. 

Though it probably warrants at least a hazarding comment about how this value will be 0 for older messages in case someone wants to repurposes this field later. I'll add one.

src/Messages/TSMessagesManager.m
michaelkirk: I don't think envelope.sourceDevice will ever be 0, but incomingMessage.sourceDevice will be 0 for models persisted before this code is deployed. Apart from commenting as such in IncomingMessage, do you still think changes are warranted here?

michaelkirk: I'll add an `assert(enevelope.sourceDevice > 0)`

src/Storage/OWSIncomingMessageFinder.m
michaelkirk: It's explained in the subsequent log message, which was originally a comment, but I figured it was more useful as a log message and silly to copy/paste it *also* as a comment. Would you have read it if the log message came before the assert?

src/Storage/OWSIncomingMessageFinder.m
michaelkirk: So, I've definitely clouded the issue by erroneously registering sync in the `asyncRegisterMessage` **THANK YOU**

But after fixing that, this code is intended for two reasons:
- race condition as you've outlined, though I'm not sure that it's possible.
- robustness against future me (or some other me) changing the app startup code, and failing to register this extension before it's used. 

Are you concerned about leaving it in?

src/Storage/OWSIncomingMessageFinder.m
michaelkirk: good catch, thank you!

src/Storage/OWSIncomingMessageFinder.m
michaelkirk: I've clarified this with an additional comment.

tests/Contacts/TSThreadTest.m
michaelkirk: done!

tests/Storage/OWSOrphanedDataCleanerTest.m
michaelkirk: done

michaelkirk:
I believe I've addressed all the outstanding concerns (and added a test)

PTAL @charlesmchen

src/Messages/TSMessagesManager.m
charlesmchen: Sounds good. I'd add a comment in `existsMessageWithTimestamp:...` that describes how the query will behave for older messages.

src/Storage/OWSIncomingMessageFinder.m
charlesmchen: Gotcha.

src/Storage/OWSIncomingMessageFinder.m
charlesmchen: Nope, makes sense, just checking.

charlesmchen:
LGTM

charlesmchen (approved):
LGTM

michaelkirk:
in https://github.com/WhisperSystems/SignalServiceKit/commit/168639597f208355b5ce1151cd45767ab8d9e6e9

PR #122, opened 2017-02-22
Add and honor the “CallKit Privacy” setting.

charlesmchen:
Adds a new TSStorageManager category to persist the "call kit id-to-phone number" mapping.

PTAL @michaelkirk

### DIFF
--- a/src/Storage/AxolotlStore/TSStorageManager+Calling.m
+++ b/src/Storage/AxolotlStore/TSStorageManager+Calling.m
@@ -4,7 +4,9 @@
 
 #import "TSStorageManager+Calling.h"
 
-#define TSStorageManagerCallKitIdToPhoneNumberCollection @"TSStorageManagerCallKitIdToPhoneNumberCollection"
+NS_ASSUME_NONNULL_BEGIN
+
+NSString *const TSStorageManagerCallKitIdToPhoneNumberCollection = @"TSStorageManagerCallKitIdToPhoneNumberCollection";
 
 @implementation TSStorageManager (Calling)
 
@@ -24,3 +26,5 @@
 }
 
 @end
+
+NS_ASSUME_NONNULL_END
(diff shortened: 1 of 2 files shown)


src/Storage/AxolotlStore/TSStorageManager+Calling.m
```diff
@@ -0,0 +1,26 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "TSStorageManager+Calling.h"
+
+#define TSStorageManagerCallKitIdToPhoneNumberCollection @"TSStorageManagerCallKitIdToPhoneNumberCollection"
```
michaelkirk: nit: I thought we decided `extern NSString *const` was the *one true way*.

src/Storage/AxolotlStore/TSStorageManager+Calling.m
```diff
@@ -0,0 +1,26 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "TSStorageManager+Calling.h"
+
```
michaelkirk, line 6: nit: I think we should use `NS_ASSUME_NONULL_BEGIN`/`END` in all objc files (.h too).

(crazy exceptions may exist, but I don't think this is one)

michaelkirk (approved):
LGTM

src/Storage/AxolotlStore/TSStorageManager+Calling.m
charlesmchen: Done.

src/Storage/AxolotlStore/TSStorageManager+Calling.m
charlesmchen: Done. Yup.  My policy is to try to clean up code as I touch it 90% of the time. This was a 10% exception since I was rushing to get the release ready before I go on vacation.

PR #123, opened 2017-03-03
Fix crash-on-launch for older installs

michaelkirk:
populating the de-duping index assumes authorId is always set for
incoming messages, but this was not always the case.

// FREEBIE

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #124, opened 2017-03-05
Charlesmchen/websocket state

charlesmchen:
I hit an edge case while testing where the socket manager learned of changes to socket state in `[ensureSocket]`, not from a delegate method. I've reorganized the socket manager to ensure that this case is handled.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/Interactions/TSInfoMessage.h
+++ b/src/Messages/Interactions/TSInfoMessage.h
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 15/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSMessage.h"
 
@@ -16,8 +17,7 @@ typedef NS_ENUM(NSInteger, TSInfoMessageType) {
     TSInfoMessageTypeDisappearingMessagesUpdate
 };
 
-+ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread
-                                     transaction:(YapDatabaseReadWriteTransaction *)transaction;
++ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread;
 
 @property TSInfoMessageType messageType;
 @property NSString *customMessage;
--- a/src/Messages/Interactions/TSInfoMessage.m
+++ b/src/Messages/Interactions/TSInfoMessage.m
@@ -45,7 +45,8 @@ NS_ASSUME_NONNULL_BEGIN
     return self;
 }
 
-+ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread {
++ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread
+{
     return [[self alloc] initWithTimestamp:[NSDate ows_millisecondTimeStamp]
                                   inThread:thread
                                messageType:TSInfoMessageUserNotRegistered];
--- a/src/Messages/Interactions/TSInfoMessage.m
+++ b/src/Messages/Interactions/TSInfoMessage.m
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 15/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "NSDate+millisecondTimeStamp.h"
 #import "TSInfoMessage.h"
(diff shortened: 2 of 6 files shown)


src/Messages/Interactions/TSInfoMessage.h
```diff
@@ -16,8 +17,7 @@ typedef NS_ENUM(NSInteger, TSInfoMessageType) {
     TSInfoMessageTypeDisappearingMessagesUpdate
 };
 
-+ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread
-                                     transaction:(YapDatabaseReadWriteTransaction *)transaction;
++ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread;
```
charlesmchen, line 20: This is some minor unrelated cleanup.  I noticed that we were needlessly passing around a transaction to this method.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -109,40 +109,132 @@ - (void)ensureWebsocket
...
 
     // Try to reuse the existing socket (if any) if it is in a valid state.
-    SRWebSocket *socket = self.websocket;
-    if (socket) {
-        switch ([socket readyState]) {
+    if (self.websocket) {
+        switch ([self.websocket readyState]) {
             case SR_OPEN:
                 DDLogVerbose(@"WebSocket already open on connection request");
-                OWSAssert(self.status == kSocketStatusOpen);
```
charlesmchen, line 117: We can't assume socket and class state are aligned in `ensureWebsocket` - indeed, we need to do the work to ensure they are aligned.

charlesmchen:
PTAL @michaelkirk

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -109,40 +109,132 @@ - (void)ensureWebsocket
...
     self.status = kSocketStatusConnecting;
+}
 
-    // Create a new web socket.
-    NSString *webSocketConnect = [textSecureWebSocketAPI stringByAppendingString:[self webSocketAuthenticationString]];
-    NSURL *webSocketConnectURL   = [NSURL URLWithString:webSocketConnect];
-    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:webSocketConnectURL];
+- (NSString *)stringFromSocketStatus:(SocketStatus)status {
+    
+    // If this status update is _not_ redundant,
```
michaelkirk: This comment seems misplaced - isn't this just a helper method to pretty-print an enum for logging purposes?

src/Messages/Interactions/TSInfoMessage.m
```diff
@@ -44,8 +45,7 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
     return self;
 }
 
-+ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread
-                                     transaction:(YapDatabaseReadWriteTransaction *)transaction {
++ (instancetype)userNotRegisteredMessageInThread:(TSThread *)thread {
```
michaelkirk: nit clang-format, brace should be on newline

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -109,40 +109,132 @@ - (void)ensureWebsocket
...
+// and [didCloseWithCode:...].  These delegate methods are sometimes
+// invoked _after_ web socket state changes, so we sometimes learn
+// about changes to socket state in [ensureWebsocket].  Put another way,
+// it's not safe to assume we'll learn of changes to websocket state
+// in the websocket delegate methods.
+//
+// Therefore, we use the [setStatus:] setter to ensure alignment between
+// websocket state and class state.
+- (void)setStatus:(SocketStatus)status {
+
```
michaelkirk: can we assert this is only called on the WS thread?

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -109,40 +109,132 @@ - (void)ensureWebsocket
...
+                                                   selector:@selector(webSocketHeartBeat)
+                                                   userInfo:nil
+                                                    repeats:YES];
+            
+            // Additionally, we want the ping timer to work in the background too.
+            [[NSRunLoop mainRunLoop] addTimer:self.pingTimer forMode:NSDefaultRunLoopMode];
+            
+            [self.reconnectTimer invalidate];
+            self.reconnectTimer = nil;
+            self.didConnectBg   = YES;
```
michaelkirk: nit: Looks like this `didConnectBg` is a vestige of a bygone era. Wanna get rid of it as long as we're tearing this method apart?

michaelkirk (approved):
LGTM

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Remove.

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Actually, it should only be called on the main thread.  I'm going to add asserts to clarify how threading works throughout this file.

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Done. I went further and removed all vestigial code in this file.

src/Messages/Interactions/TSInfoMessage.m
charlesmchen: Done.

PR #125, opened 2017-03-11
Add UTIType methods to MIMETypeUtil.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
LGTM 

Think it's worth memoizing these?

charlesmchen:
> Think it's worth memoizing these?

I've memoized _all_ of the collections returned by `MIMETypeUtils`.

PR #126, opened 2017-03-13
Log asserts in production.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
To catch issues like the "debug logs not working" issue, I'm thinking of modifying our assert macros to log using NSLog in production.  

* This won't affect user experience.
* This does mean that asserts will be evaluated, which implies a slight perf cost - but Signal doesn't seem to be an app where perf is much of a concern for 99% of the codebase. We could add a OWSDebugAssert() for debug-only asserts if we needed to.
* By logging using NSLog, the logs will only show up in the console, not in the debug logs so we won't need to worry about the debug logs capturing private user info.

PR #127, opened 2017-03-13
Attempt to fix the "frequent attachment download errors with low serv…

charlesmchen:
…er ids".

I'll explain this work inline.

PTAL @michaelkirk

### DIFF
--- a/src/Devices/OWSReadReceiptsProcessor.m
+++ b/src/Devices/OWSReadReceiptsProcessor.m
@@ -1,4 +1,6 @@
-//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "OWSReadReceiptsProcessor.h"
 #import "OWSDisappearingMessagesJob.h"
@@ -91,7 +93,7 @@ NSString *const OWSReadReceiptsProcessorMarkedMessageAsReadNotification =
                               object:message];
 
         } else {
-            DDLogDebug(@"%@ Received read receipt for an unkown message. Saving it for later.", self.tag);
+            DDLogDebug(@"%@ Received read receipt for an unknown message. Saving it for later.", self.tag);
             [readReceipt save];
         }
     }
--- a/src/Messages/Attachments/OWSAttachmentsProcessor.m
+++ b/src/Messages/Attachments/OWSAttachmentsProcessor.m
@@ -119,6 +119,9 @@ NS_ASSUME_NONNULL_BEGIN
         }
     };
 
+    if (attachment.serverId < 100) {
+        DDLogError(@"%@ Suspicious attachment id: %llu", self.tag, (unsigned long long)attachment.serverId);
+    }
     TSAttachmentRequest *attachmentRequest = [[TSAttachmentRequest alloc] initWithId:attachment.serverId relay:attachment.relay];
 
     [self.networkManager makeRequest:attachmentRequest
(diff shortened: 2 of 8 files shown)


src/Devices/OWSReadReceiptsProcessor.m
```diff
@@ -91,7 +93,7 @@ - (void)process
                               object:message];
 
         } else {
-            DDLogDebug(@"%@ Received read receipt for an unkown message. Saving it for later.", self.tag);
+            DDLogDebug(@"%@ Received read receipt for an unknown message. Saving it for later.", self.tag);
```
charlesmchen, line 96: Just a typo fix.

src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -118,6 +118,11 @@ - (void)retrieveAttachment:(TSAttachmentPointer *)attachment
         }
     };
 
+    if (attachment.serverId < 100) {
+        DDLogError(
```
charlesmchen: I've modified the attachment download logic to be more verbose around the symptom.

src/Messages/Attachments/TSAttachment.m
```diff
@@ -33,18 +34,34 @@ - (instancetype)initWithServerId:(UInt64)serverId
...
+}
+
 - (nullable instancetype)initWithCoder:(NSCoder *)coder
 {
     self = [super initWithCoder:coder];
     if (!self) {
         return self;
     }
 
-    if (_attachmentSchemaVersion < 2) {
```
charlesmchen: This is the heart of the possible fix.  

* We were doing a schema upgrade in `TSAttachment`, which is the base class of `TSAttachmentPointer` and `TSAttachmentStream`.  
* For any instance of `TSAttachmentStream`, this upgrade was inappropriate and possibly lead to data loss. Instances of `TSAttachmentStream` don't set their attachmentSchemaVersion, so they were always be "upgraded".  
* `TSAttachmentStream` represents two separate things: an outgoing attachment or an already-downloaded incoming attachment.  I don't see how any harm could be done by the latter case, but the former case is dangerous.
* An outgoing attachment's server id is set after the upload is complete.  If the `TSAttachmentStream` was reloaded before the corresponding message is sent to the service, that message will end up with an incorrect service id.
* The data loss occurs because the schema upgrade overwrites the service id by parsing the uniqueId of the attachment.  For any new attachment, the uniqueId will be a UUID.  The uniqueId is parsed using `[NSString integerValue]`.  Given a UUID value like 11BFAA01-ADD4-4E48-81B2-2654941C9ECD, this method will simply parse the prefix characters that can be parsed as an integer, in this case the prefix 11.  This would explain why the 404s so frequently have small server ids.
* `TSOutgoingMessage` doesn't store instances of `TSAttachmentStream` - it stores the uniqueIds for the message's attachments.  This would explain how the `TSAttachmentStream` would get reloaded between attachment upload completion and message upload.
* So it all fits pretty neatly.  The only problem is that even with this diagnosis I can't find a repro.  Nonetheless the current behavior was clearly wrong so we should merge this change.

src/Messages/Attachments/TSAttachmentStream.h
```diff
@@ -16,8 +17,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithContentType:(NSString *)contentType NS_DESIGNATED_INITIALIZER;
 - (instancetype)initWithPointer:(TSAttachmentPointer *)pointer NS_DESIGNATED_INITIALIZER;
 
-@property (atomic, readwrite) BOOL isDownloaded;
```
charlesmchen, line 20: I've promoted this property to the base class `TSAttachment` so that we can use it to determine if we need to apply the schema upgrade.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -130,6 +130,8 @@ - (void)ensureWebsocket
     // Discard the old socket which is already closed or is closing.
     [self closeWebSocket];
 
+    DDLogWarn(@"Creating new websocket");
```
charlesmchen: I've seen a bit of jankiness around the websocket lifecycle that might just be edge cases in the simulator - or might not.  Since having a valid websocket is critical, I've added extra logging.

src/Messages/Attachments/TSAttachment.m
michaelkirk: Ack! Good find.

I agree with your assessment of what is broken, and I think this code fixes it.

Regardless of this fix, we should be properly setting the TSAttachmentSchemaVersion for subclasses of TSAttachment (e.g. with designated initializers in this class), else every future migration will likely suffer the same problem.

And if we're going to do that, I think we could/should remove these changes.

whaddayathink?

charlesmchen:
PTAL @michaelkirk

michaelkirk (commented):
Ack, forgot to submit my review from 2 days ago.

src/Messages/Attachments/TSAttachment.m
charlesmchen: > Regardless of this fix, we should be properly setting the TSAttachmentSchemaVersion for subclasses of TSAttachment (e.g. with designated initializers in this class), else every future migration will likely suffer the same problem.

I've modified `TSAttachmentStream` to use initializers in `TSAttachment` so that `TSAttachmentSchemaVersion` is set for new instances.  I'm using these new initializers to ensure that new  `TSAttachmentStream` that represented downloaded incoming attachments no longer discard their server id and encryption key, with an eye towards supporting re-download in the future.

> And if we're going to do that, I think we could/should remove these changes.

These changes are still necessary, because`TSAttachmentSchemaVersion` still won't be set for legacy instances of `TSAttachmentStream`.  I'm going to simplify things and leverage the changes made in https://github.com/WhisperSystems/SignalServiceKit/pull/129 and move this to `TSAttachmentPointer`.

charlesmchen:
I've significantly reworked this PR.

PTAL @michaelkirk

src/Messages/Attachments/TSAttachmentStream.h
michaelkirk: nit: It seems like we're no longer using it to determine the migration logic. 

In fact, unless I'm missing a condition, it seems like we could avoid redundant state tracking and do:

```
@implementation TSAttachment

@property (nonatomic, readonly) BOOL isDownloaded;

@end 

@implementation TSAttachmentStream
- (BOOL)isDownloaded
{
    return YES;
}
@end

@implementation TSAttachmentPointer
- (BOOL)isDownloaded
{
    return NO;
}
@end

michaelkirk (approved):
LGTM unless my nit has persuaded you...

src/Messages/Attachments/TSAttachmentStream.h
charlesmchen: Once we merge https://github.com/WhisperSystems/Signal-iOS/pull/1849 we should be able to do away with this property altogether. Sound like a plan?

PR #128, opened 2017-03-14
Fix filtering of country codes in registration flow.

charlesmchen:
See https://github.com/WhisperSystems/Signal-iOS/pull/1835

PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

michaelkirk (approved):
LGTM

PR #129, opened 2017-03-15
Add "is uploaded" property to attachment streams.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/TSAttachmentStream.m
+++ b/src/Messages/Attachments/TSAttachmentStream.m
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 17/12/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSAttachmentStream.h"
 #import "MIMETypeUtil.h"
@@ -20,6 +21,10 @@ NS_ASSUME_NONNULL_BEGIN
 
     _contentType = contentType;
     _isDownloaded = YES;
+    // TSAttachmentStream doesn't have any "incoming vs. outgoing"
+    // state, but this constructor is used only for new outgoing
+    // attachments which haven't been uploaded yet.
+    _isUploaded = NO;
 
     return self;
 }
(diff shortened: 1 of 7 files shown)


src/Network/API/OWSUploadingService.m
```diff
@@ -79,9 +83,11 @@ - (void)uploadAttachmentStream:(TSAttachmentStream *)attachmentStream
                                     location:location
                                 attachmentId:attachmentStream.uniqueId
                                      success:^{
-                                         DDLogInfo(@"%@ Uploaded attachment.", self.tag);
+                                         OWSAssert([NSThread isMainThread]);
+
+                                         DDLogInfo(@"%@ Uploaded attachment: %p.", self.tag, attachmentStream);
                                          attachmentStream.serverId = serverId;
-                                         attachmentStream.isDownloaded = YES;
```
michaelkirk, line 84: I guess this was redundant since it's already set to YES in the initializer?

src/Messages/Interactions/TSOutgoingMessage.h
```diff
@@ -12,9 +12,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+    // The message is either:
+    // a) Enqueued for sending.
+    // b) Waiting on attachment upload(s).
+    // c) Being sent to the service.
     TSOutgoingMessageStateAttemptingOut,
+    // The failure state.
     TSOutgoingMessageStateUnsent,
+    // The message has been sent to the service, but not received by any recipient client.
     TSOutgoingMessageStateSent,
+    // The message has been sent to the service, but not received by at least one recipient client.
```
michaelkirk: I think you mean:

> but not received by at least one recipient client.
~~but not~~ *and* received by at least one recipient client.

michaelkirk (approved):
LGTM

src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -34,10 +39,27 @@ - (instancetype)initWithPointer:(TSAttachmentPointer *)pointer
...
     _isDownloaded = YES;
+    // TSAttachmentStream doesn't have any "incoming vs. outgoing"
+    // state, but this constructor is used only for new incoming
+    // attachments which don't need to be uploaded.
+    _isUploaded = YES;
 
     return self;
 }
 
+- (void)upgradeFromAttachmentSchemaVersion:(NSUInteger)attachmentSchemaVersion
```
charlesmchen, line 50: Pulling out the schema version upgrade logic to a method so that subclasses can override and add upgrade logic of their own.

michaelkirk: Yeah - it never occurred to me to do this, but I think I like it.

Now I'm wondering if we should just have a unified schema version across all our TSYapDatabaseObjects instead of tracking per model 👀 ...

src/Messages/Interactions/TSOutgoingMessage.h
charlesmchen: Done.

src/Network/API/OWSUploadingService.m
charlesmchen: Replied offline.

PR #130, opened 2017-03-15
Fix non-contact lookup for non-US users.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/ContactsUpdater.h
+++ b/src/Contacts/ContactsUpdater.h
@@ -14,10 +14,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (nullable SignalRecipient *)synchronousLookup:(NSString *)identifier error:(NSError **)error;
 
-// This asynchronously updates the SignalRecipient for a given contactId.
+// This asynchronously tries to verify whether or not a contact id
+// corresponds to a service account.
+//
+// The failure callback is invoked if the lookup fails _or_ if the
+// contact id doesn't correspond to an account.
 - (void)lookupIdentifier:(NSString *)identifier
                  success:(void (^)(SignalRecipient *recipient))success
                  failure:(void (^)(NSError *error))failure;
+
+// This asynchronously tries to verify whether or not group of possible
+// contact ids correspond to service accounts.
+//
+// The failure callback is only invoked if the lookup fails.  Otherwise,
+// the success callback is invoked with the (possibly empty) set of contacts
+// that were found.
+- (void)lookupIdentifiers:(NSArray<NSString *> *)identifiers
+                  success:(void (^)(NSArray<SignalRecipient *> *recipients))success
+                  failure:(void (^)(NSError *error))failure;
 
 - (void)updateSignalContactIntersectionWithABContacts:(NSArray<Contact *> *)abContacts
                                               success:(void (^)())success
--- a/src/Contacts/ContactsUpdater.h
+++ b/src/Contacts/ContactsUpdater.h
@@ -23,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
                  success:(void (^)(SignalRecipient *recipient))success
                  failure:(void (^)(NSError *error))failure;
 
-// This asynchronously tries to verify whether or not group of possible
+// This asynchronously tries to verify whether or not a group of possible
 // contact ids correspond to service accounts.
 //
 // The failure callback is only invoked if the lookup fails.  Otherwise,
--- a/src/Contacts/ContactsUpdater.m
+++ b/src/Contacts/ContactsUpdater.m
@@ -60,11 +60,36 @@ NS_ASSUME_NONNULL_BEGIN
         failure(OWSErrorWithCodeDescription(OWSErrorCodeInvalidMethodParameters, @"Cannot lookup nil identifier"));
         return;
     }
-
+    
     [self contactIntersectionWithSet:[NSSet setWithObject:identifier]
                              success:^(NSSet<NSString *> *_Nonnull matchedIds) {
                                  if (matchedIds.count == 1) {
                                      success([SignalRecipient recipientWithTextSecureIdentifier:identifier]);
+                                 } else {
+                                     failure(OWSErrorMakeNoSuchSignalRecipientError());
+                                 }
+                             }
+                             failure:failure];
+}
+
+- (void)lookupIdentifiers:(NSArray<NSString *> *)identifiers
+                 success:(void (^)(NSArray<SignalRecipient *> *recipients))success
+                 failure:(void (^)(NSError *error))failure
+{
+    if (identifiers.count < 1) {
+        OWSAssert(NO);
+        failure(OWSErrorWithCodeDescription(OWSErrorCodeInvalidMethodParameters, @"Cannot lookup zero identifiers"));
+        return;
+    }
+    
+    [self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
+                             success:^(NSSet<NSString *> *_Nonnull matchedIds) {
+                                 if (matchedIds.count == 1) {
+                                     NSMutableArray<SignalRecipient *> *recipients = [NSMutableArray new];
+                                     for (NSString *identifier in matchedIds) {
+                                         [recipients addObject:[SignalRecipient recipientWithTextSecureIdentifier:identifier]];
+                                     }
+                                     success(recipients);
                                  } else {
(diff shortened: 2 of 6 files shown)


charlesmchen:
Wait, I thought of a way to improve this.

charlesmchen:
Okay, I've updated this PR. PTAL @michaelkirk

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1852

src/Contacts/ContactsUpdater.h
```diff
@@ -14,11 +14,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+// This asynchronously tries to verify whether or not a contact id
+// corresponds to a service account.
+//
+// The failure callback is invoked if the lookup fails _or_ if the
+// contact id doesn't correspond to an account.
 - (void)lookupIdentifier:(NSString *)identifier
                  success:(void (^)(SignalRecipient *recipient))success
                  failure:(void (^)(NSError *error))failure;
 
+// This asynchronously tries to verify whether or not group of possible
```
michaelkirk: nit
> This asynchronously tries to verify whether or not group of possible

This asynchronously tries to verify whether or not **a** group of possible

src/Contacts/ContactsUpdater.m
```diff
@@ -72,6 +72,31 @@ - (void)lookupIdentifier:(NSString *)identifier
...
+    }
+    
+    [self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
+                             success:^(NSSet<NSString *> *_Nonnull matchedIds) {
+                                 if (matchedIds.count == 1) {
+                                     NSMutableArray<SignalRecipient *> *recipients = [NSMutableArray new];
+                                     for (NSString *identifier in matchedIds) {
+                                         [recipients addObject:[SignalRecipient recipientWithTextSecureIdentifier:identifier]];
+                                     }
+                                     success(recipients);
```
michaelkirk: nit: `[recipients copy]` to return immutable

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
     return [self phoneNumberFromUserSpecifiedText:sanitizedString];
 }
 
++ (NSArray *)tryParsePhoneNumbersFromsUserSpecifiedText:(NSString *)text
```
michaelkirk: can you type annotate this array - I think it's a `NSArray<PhoneNumber *> *` right? (and in the .h)

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
+                                      clientPhoneNumber:(NSString *)clientPhoneNumber {
+    assert(text != nil);
```
michaelkirk: OWSAssert? NSAssert? assert? Who am I? What day is it?

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
+    assert(sanitizedString != nil);
+    
+    NSMutableArray *result = [NSMutableArray new];
+    NSMutableSet *phoneNumberSet = [NSMutableSet new];
+    void (^tryParsingWithCountryCode)(NSString *, NSString *) = ^(NSString *text,
+                                                      NSString *countryCode) {
+        PhoneNumber *phoneNumber = [PhoneNumber phoneNumberFromText:text
+                                                          andRegion:countryCode];
+        if (phoneNumber &&
+            ![phoneNumberSet containsObject:[phoneNumber toE164]]) {
```
michaelkirk: nit: This set+array code might be a little more straight forward with a dict

```
if (phoneNumber) {
    phoneNumbers[phoneNumber.toE164] = phoneNumber;
}

```

It's unclear to me if`toE164` can ever return nil. Maybe to be on the safe side...

```
if (phoneNumber && phoneNumber.toE164) {
    phoneNumbers[phoneNumber.toE164] = phoneNumber;
}

```

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
+
+    if ([sanitizedString hasPrefix:@"+"]) {
+        // If the text starts with "+", just parse it.
+        tryParsingWithCountryCode(sanitizedString, [self defaultRegionCode]);
+
+        return result;
+    }
+
+    // Order matters; better results should appear first so
+    // prefer matches using this client's phone number.
```
michaelkirk: clarity nit: // prefer matches ~~using~~ **with the same country code as** this client's phone number.

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
+    if (clientPhoneNumber.length > 0) {
```
michaelkirk: Do we ever expect the client not to have a phone number? Should this be an assert and/or guard to communicate that?

src/Contacts/PhoneNumber.m
```diff
@@ -115,6 +119,75 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
+        NSNumber *callingCodeForLocalNumber = [[PhoneNumber phoneNumberFromE164:clientPhoneNumber] getCountryCode];
+        if (callingCodeForLocalNumber != nil) {
+            tryParsingWithCountryCode([NSString stringWithFormat:@"+%@%@",
+                                       callingCodeForLocalNumber,
+                                       sanitizedString],
+                                      [self defaultRegionCode]);
+
+            if ([sanitizedString hasPrefix:callingCodeForLocalNumber.description]) {
```
michaelkirk: Is this an idiomatic way to convert an NSNumber to a string?

charlesmchen:
PTAL @michaelkirk

michaelkirk (approved):
LGTM

src/Contacts/ContactsUpdater.h
charlesmchen: Done.

src/Contacts/ContactsUpdater.m
charlesmchen: Done.

src/Contacts/PhoneNumber.m
charlesmchen: Done.

src/Contacts/PhoneNumber.m
charlesmchen: Done.

src/Contacts/PhoneNumber.m
charlesmchen: We can't used a dict because we want to return the `PhoneNumber`s and the ordering of the results matters. a) we want the ordering to be stable to provide a predictable UX b) we want "better" matches to appear earlier in the results.

I've added a check that `toE164` is non-nil.

src/Contacts/PhoneNumber.m
charlesmchen: I tweaked your suggestion.

src/Contacts/PhoneNumber.m
michaelkirk: >We can't used a dict because we want to return the PhoneNumbers and the ordering of the results matters.

Oh right - my bad.

src/Contacts/PhoneNumber.m
charlesmchen: Well, I was trying to leave this open to be used when the app is non-registered but I guess I'll add an assert for now.

src/Contacts/PhoneNumber.m
charlesmchen: Well, using a NSNumberFormatter or `[NSString stringWithFormat:@"%d", someInt]` is so cumbersome that for simple values (in this case, a positive, small int) I'll often use `description`.  It's a borderline abuse, but it's exactly what happens if you do `[NSString stringWithFormat:@"%@", @(someInt)]`.  

Anyway, I'll switch to the NSString approach, but it's cumbersome-ness irks me.

PR #131, opened 2017-03-17
Serialize method sending

### LINKED ISSUE #1840: We need to improve effort to ensure message ordering.
See screenshots.

Repro 1:

* Send attachment A.
* Send message B.

Expected: A will be delivered before B.
Observed: B will typically be delivered before A.

___We should enqueue messages in the order in which they are sent and upload them in serial order - unless a given message exhausts its retries and fails.___

Repro 2:

* Enable network link conditioner in "very bad network" mode (which will cause frequent network errors).  It's worth noting that this mode reflects common network conditions for mobile users.
* Send messages A, B, C, D, E, F, G in rapid sequence (should each have a different, very short text body).

Expected: A, B, C, etc. will be delivered in order.
Observed: Messages will arrive in nearly random order.

___We should enqueue messages in the order in which they are sent and upload them in serial order - unless a given message exhausts its retries and fails.___

### Sender

img_4298

### Recipient

screen shot 2017-03-12 at 6 45 47 pm

michaelkirk:
fixes https://github.com/WhisperSystems/Signal-iOS/issues/1840

Also added retries that cover attachment failures
Reduced latency before showing user outgoing message bubbles for a snappier UX

PTAL @charlesmchen

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -246,6 +246,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             success:(void (^)())successHandler
             failure:(void (^)(NSError *error))failureHandler
 {
+    [self saveMessage:message withState:TSOutgoingMessageStateAttemptingOut];
     OWSSendMessageOperation *sendMessageOperation = [[OWSSendMessageOperation alloc] initWithMessage:message
                                                                                        messageSender:self
                                                                                              success:successHandler
@@ -339,8 +340,6 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 
         [attachmentStream save];
         [message.attachmentIds addObject:attachmentStream.uniqueId];
-
-        message.messageState = TSOutgoingMessageStateAttemptingOut;
         [message save];
 
         [self sendMessage:message success:successHandler failure:failureHandler];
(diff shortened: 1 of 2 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -41,6 +41,163 @@
...
+    _messageSender = messageSender;
+
+    __weak typeof(self) weakSelf = self;
+    _successHandler = ^{
+        typeof(self) strongSelf = weakSelf;
+        if (!strongSelf) {
+            return;
+        }
+        DDLogDebug(@"%@ succeeded.", strongSelf.tag);
+        aSuccessHandler();
```
charlesmchen, line 115: nit: Could the "caller" and "internal" succuess/failure handlers have names that better disambiguated them?  No need to do anything.

src/Messages/OWSMessageSender.m
```diff
@@ -41,6 +41,163 @@
...
+    }
+
+    _operationState = OWSSendMessageOperationStateNew;
+
+    _message = message;
+    _messageSender = messageSender;
+
+    __weak typeof(self) weakSelf = self;
+    _successHandler = ^{
+        typeof(self) strongSelf = weakSelf;
```
charlesmchen, line 110: Shouldn't we assert strongSelf in these handlers?

src/Messages/OWSMessageSender.m
```diff
@@ -41,6 +41,163 @@
...
+
+@property (nonatomic, readonly) TSOutgoingMessage *message;
+@property (nonatomic, readonly) OWSMessageSender *messageSender;
+@property (nonatomic, readonly) void (^successHandler)();
+@property (nonatomic, readonly) void (^failureHandler)(NSError *_Nonnull error);
+@property (atomic) OWSSendMessageOperationState operationState;
+
+@end
+
+@implementation OWSSendMessageOperation
```
charlesmchen, line 91: It looks like this operation assume it will never be cancelled, because if it is cancelled before it is started, no handler will ever be called.  I think that's fine.

charlesmchen (approved):
LGTM

michaelkirk:
In https://github.com/WhisperSystems/SignalServiceKit/commit/15f5c078a19af169983409dd288f1db068a32eab

src/Messages/OWSMessageSender.m
michaelkirk: done

PR #132, opened 2017-03-17
Mkirk/background sending

michaelkirk:
based on https://github.com/WhisperSystems/SignalServiceKit/pull/131

Adds support to make sure we don't suspend the app with an inflight message.

PTAL @charlesmchen


charlesmchen (approved):
LGTM

michaelkirk:
in 778d3dd

PR #133, opened 2017-03-17
Fix receiving attachments from old clients

michaelkirk:
Introduced with the new digesting. Protobuf's null-object pattern provides an empty NSData when no digest is provided rather than nil.

Thought I tested this, but obviously I did not  =/ 
(test doc updated)

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #135, opened 2017-03-20
Only call sendMessage on main thread.

michaelkirk:
fixup to #132 

The proximate cause for this change was a failing assert in the recently
introduced call to `setBackgroundTaskIdentifier` which is synchronized
on the main thread.

But generally, to keep things simple, we prefer calling back on main
thread unless performance dictates otherwise.

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #136, opened 2017-03-20
Abort attachment downloads of excessive size.

charlesmchen:
This achieves parity with the Android client's behavior.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/OWSAttachmentsProcessor.m
+++ b/src/Messages/Attachments/OWSAttachmentsProcessor.m
@@ -233,22 +233,64 @@ NS_ASSUME_NONNULL_BEGIN
     manager.responseSerializer = [AFHTTPResponseSerializer serializer];
     manager.completionQueue    = dispatch_get_main_queue();
 
+    // We want to avoid large downloads from a compromised or buggy service.
+    const long kMaxDownloadSize = 150 * 1024 * 1024;
     // TODO stream this download rather than storing the entire blob.
-    [manager GET:location
-      parameters:nil
-        progress:nil // TODO show some progress!
-         success:^(NSURLSessionDataTask *_Nonnull task, id _Nullable responseObject) {
-             if (![responseObject isKindOfClass:[NSData class]]) {
-                 DDLogError(@"%@ Failed retrieval of attachment. Response had unexpected format.", self.tag);
-                 NSError *error = OWSErrorMakeUnableToProcessServerResponseError();
-                 return failureHandler(task, error);
-             }
-             successHandler((NSData *)responseObject);
-         }
-         failure:^(NSURLSessionDataTask *_Nullable task, NSError *_Nonnull error) {
-             DDLogError(@"Failed to retrieve attachment with error: %@", error.description);
-             return failureHandler(task, error);
-         }];
+    __block NSURLSessionDataTask *task = nil;
+    // We only need to check the content length header once.
+    __block BOOL hasCheckedContentLength = NO;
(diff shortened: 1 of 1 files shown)


src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -233,22 +233,64 @@ - (void)downloadFromLocation:(NSString *)location
...
+                                if (contentLength.longLongValue <= kMaxDownloadSize) {
+                                    // This response has a valid content length that is less
+                                    // than our max download size.  Proceed with the download.
+                                    hasCheckedContentLength = YES;
+                                    return;
+                                }
+                            }
+                        }
+                    }
+                }
```
michaelkirk: nit: This is nested pretty deep, to the point that I guessed wrong twice when trying to eyeball which condition this bracket closes. I think the above could be made much clearer by using guard statements and some kind of early return (in this case probably utilizing a block that has an assertion failure and a `[task cancel]`).

Also it would allow you to log a specific error at the point of the failing test case. =)

src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -233,22 +233,64 @@ - (void)downloadFromLocation:(NSString *)location
...
-         failure:^(NSURLSessionDataTask *_Nullable task, NSError *_Nonnull error) {
-             DDLogError(@"Failed to retrieve attachment with error: %@", error.description);
-             return failureHandler(task, error);
-         }];
+    __block NSURLSessionDataTask *task = nil;
+    // We only need to check the content length header once.
+    __block BOOL hasCheckedContentLength = NO;
+    task = [manager GET:location
+        parameters:nil
+        progress:^(NSProgress *_Nonnull progress) {
```
michaelkirk: nit: I'm not sure what kind of guarantees we have about when `progress` gets called. Probably it gets called at least once? But from what I can see, the NSURLSessionDataDelegate is *the way*™; to inspect the headers of a cancelable data task. 

https://developer.apple.com/reference/foundation/nsurlsessiondatadelegate/1410027-urlsession?language=objc

Also it looks like using the NSURLSessionDataDelegate would pave the way to one day graduate to a proper file download task instead of a data task, which I *think* would allow us to safely download larger attachments in the background *and* allow us to pass around streams/file handles instead of large blobs of binary data. 🎉  

WDYT?

michaelkirk (approved):

src/Messages/Attachments/OWSAttachmentsProcessor.m
charlesmchen: I'll document our offline conversation here.

> Also it looks like using the NSURLSessionDataDelegate would pave the way to one day graduate to a proper file download task instead of a data task, which I think would allow us to safely download larger attachments in the background and allow us to pass around streams/file handles instead of large blobs of binary data. 🎉

Yes, but this would be a bigger project than just adding a delegate since tasks with delegates no longer call their success & failure handlers IIRC.  I think this would take a developer day, so I'm punting.  I've created an issue.

https://github.com/WhisperSystems/Signal-iOS/issues/1872

src/Messages/Attachments/OWSAttachmentsProcessor.m
charlesmchen: Done. I deliberately avoided this to avoid tripling the size of this logic, but you're right - logging each possible failure separately will be helpful.

PR #137, opened 2017-03-21
Fix edge cases around the websocket lifecycle.

charlesmchen:
Found even more issues around the websocket lifecycle.

PTAL @michaelkirk

### DIFF
--- a/src/Network/WebSockets/TSSocketManager.m
+++ b/src/Network/WebSockets/TSSocketManager.m
@@ -160,6 +160,10 @@ NSString *const SocketConnectingNotification = @"SocketConnectingNotification";
 
     // If this status update is redundant, verify that
     // class state and socket state are aligned.
+    //
+    // Note: it's not safe to check the socket's readyState here as
+    //       it may have been just updated on another thread. If so,
+    //       we'll learn of that state change soon.
     if (_status == status) {
         switch (status) {
             case kSocketStatusClosed:
@@ -167,11 +171,9 @@ NSString *const SocketConnectingNotification = @"SocketConnectingNotification";
                 break;
             case kSocketStatusOpen:
                 OWSAssert(self.websocket);
-                OWSAssert([self.websocket readyState] == SR_OPEN);
                 break;
             case kSocketStatusConnecting:
                 OWSAssert(self.websocket);
-                OWSAssert([self.websocket readyState] == SR_CONNECTING);
                 break;
         }
         return;
--- a/src/Network/WebSockets/TSSocketManager.m
+++ b/src/Network/WebSockets/TSSocketManager.m
@@ -57,7 +57,8 @@ NSString *const SocketConnectingNotification = @"SocketConnectingNotification";
     OWSAssert([NSThread isMainThread]);
 
     _signalService = [OWSSignalService new];
-    
+    _status = kSocketStatusClosed;
+
     [self addObserver:self forKeyPath:@"status" options:0 context:kSocketStatusObservationContext];
 
     return self;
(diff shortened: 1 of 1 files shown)


src/Network/WebSockets/TSSocketManager.m
```diff
@@ -57,7 +57,8 @@ - (instancetype)init
     OWSAssert([NSThread isMainThread]);
 
     _signalService = [OWSSignalService new];
-    
+    _status = kSocketStatusClosed;
```
charlesmchen, line 60: We were neglecting to set this to a valid initial value.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -117,6 +118,13 @@ - (void)ensureWebsocket
     DDLogWarn(@"Creating new websocket");
 
     // If socket is not already open or connecting, connect now.
+    //
+    // First we need to close the existing websocket, if any.
+    // The websocket methods are invoked _after_ the websocket state
+    // changes, so we may be just learning about a socket failure
+    // or close event now.
+    self.status = kSocketStatusClosed;
```
charlesmchen: This is a key fix, as it ensures that we clear out any old failed/closed web sockets.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -152,18 +160,20 @@ - (void)setStatus:(SocketStatus)status {
...
             case kSocketStatusClosed:
                 OWSAssert(!self.websocket);
                 break;
             case kSocketStatusOpen:
                 OWSAssert(self.websocket);
-                OWSAssert([self.websocket readyState] == SR_OPEN);
                 break;
             case kSocketStatusConnecting:
                 OWSAssert(self.websocket);
-                OWSAssert([self.websocket readyState] == SR_CONNECTING);
```
charlesmchen, line 166: This wasn't safe - see comment above.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -211,7 +221,13 @@ - (void)setStatus:(SocketStatus)status {
             socket.delegate = self;
             
             [self setWebsocket:socket];
+
+            // [SRWebSocket open] could hypothetically call a delegate method (e.g. if
+            // the socket failed immediately for some reason), so we update the status
+            // _before_ calling it, not after.
+            _status = status;
             [socket open];
+            return;
```
charlesmchen, line 230: I don't think this is necessary in practice, but this isn't the first time we've hit edge cases with our web socket, so I'm trying to be super defensive.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -250,12 +266,22 @@ - (void)closeWebSocket
 
 - (void)webSocketDidOpen:(SRWebSocket *)webSocket {
     OWSAssert([NSThread isMainThread]);
-    
+    OWSAssert(webSocket);
+    if (webSocket != self.websocket) {
+        // Ignore events from obsolete web sockets.
+        return;
+    }
```
charlesmchen, line 273: This (and the corresponding changes below) does turn out to be necessary.  Apparently the websocket captures it's delegate _before_ dispatching the delegate method invocation.

charlesmchen:
The repro around these issues is just to launch while the phone has no valid internet connection.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -117,6 +118,13 @@ - (void)ensureWebsocket
     DDLogWarn(@"Creating new websocket");
 
     // If socket is not already open or connecting, connect now.
+    //
+    // First we need to close the existing websocket, if any.
+    // The websocket methods are invoked _after_ the websocket state
```
michaelkirk: By "the websocket methods" do you mean "the websocket delegate methods"?

michaelkirk (approved):
LGTM

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Done.

PR #138, opened 2017-03-21
Improve logging around decryption exceptions.

charlesmchen:
Working on the "Recipients key is not valid" issue.

PTAL @michaelkirk


michaelkirk (approved):
LGTM

PR #139, opened 2017-03-21
Further refinements to phone number parsing.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/PhoneNumber.m
+++ b/src/Contacts/PhoneNumber.m
@@ -144,30 +144,31 @@ static NSString *const RPDefaultsKeyPhoneNumberCanonical = @"RPDefaultsKeyPhoneN
         }
     };
 
-    if ([sanitizedString hasPrefix:@"+"]) {
-        // If the text starts with "+", just parse it.
-        tryParsingWithCountryCode(sanitizedString, [self defaultRegionCode]);
+    tryParsingWithCountryCode(sanitizedString, [self defaultRegionCode]);
 
+    if ([sanitizedString hasPrefix:@"+"]) {
+        // If the text starts with "+", don't try prepending
+        // anything else.
         return result;
     }
+
+    // Try just adding "+" and parsing it.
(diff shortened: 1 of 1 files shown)


src/Contacts/PhoneNumber.m
```diff
@@ -144,13 +144,17 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
         }
     };
 
-    if ([sanitizedString hasPrefix:@"+"]) {
-        // If the text starts with "+", just parse it.
-        tryParsingWithCountryCode(sanitizedString, [self defaultRegionCode]);
+    tryParsingWithCountryCode(sanitizedString, [self defaultRegionCode]);
```
charlesmchen, line 147: I decided the ordering was wrong; this result should be first, not last.

src/Contacts/PhoneNumber.m
```diff
@@ -144,13 +144,17 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
...
 
+    if ([sanitizedString hasPrefix:@"+"]) {
+        // If the text starts with "+", don't try prepending
+        // anything else.
         return result;
     }
 
+    // Try just adding "+" and parsing it.
+    tryParsingWithCountryCode([NSString stringWithFormat:@"+%@", sanitizedString], [self defaultRegionCode]);
```
charlesmchen, line 156: The key fix: We should always try parsing by adding "+" regardless of whether it starts with the calling code that corresponds to the device's default region.

charlesmchen:
The repro is: 

* Use a device registered with a +1 (US) phone number.
* Enter a +44 (UK) E164 phone number ___but without___ a "+" plus sign that corresponds to a signal account. 

The phone number should show up as a signal user.

michaelkirk (approved):
LGTM

PR #142, opened 2017-03-23
Use a separate sending queue for each conversation.

charlesmchen:
I was surprised to see that there are outgoing interactions for which thread and uniqueThreadId are nil.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -257,7 +257,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 @property (nonatomic, readonly) id<ContactsManagerProtocol> contactsManager;
 @property (nonatomic, readonly) ContactsUpdater *contactsUpdater;
 @property (nonatomic, readonly) OWSDisappearingMessagesJob *disappearingMessagesJob;
-@property (nonatomic, readonly) NSOperationQueue *sendingQueue;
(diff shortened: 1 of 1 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -257,7 +257,7 @@ @interface OWSMessageSender ()
 @property (nonatomic, readonly) id<ContactsManagerProtocol> contactsManager;
 @property (nonatomic, readonly) ContactsUpdater *contactsUpdater;
 @property (nonatomic, readonly) OWSDisappearingMessagesJob *disappearingMessagesJob;
-@property (nonatomic, readonly) NSOperationQueue *sendingQueue;
+@property (atomic, readonly) NSMutableDictionary *sendingQueueMap;
```
michaelkirk: nit: can you type annotate this dictionary.

And why can't it be nonatomic? We're only assigning it in the initializer.

michaelkirk (approved):
LGTM

src/Messages/OWSMessageSender.m
charlesmchen: Annotate: done.

atomic: Although its read-only access today, I like use atomic for any property that might be accessed from multiple threads.  It's an attempt to improve readability.  I might think twice about it if this was a perf hotspot, but it's very far from being that.

PR #147, opened 2017-03-26
Filter out country codes properly.

charlesmchen:
We always need to filter the country codes, even when fetching all.

PTAL @michaelkirk

### DIFF
--- a/src/Contacts/PhoneNumberUtil.m
+++ b/src/Contacts/PhoneNumberUtil.m
@@ -117,33 +117,36 @@
 
     NSArray *countryCodes = NSLocale.ISOCountryCodes;
 
-    if (searchTerm.length > 0) {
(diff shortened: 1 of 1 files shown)


michaelkirk (approved):
LGTM

What is the user facing behavior this fixes?

charlesmchen:
@michaelkirk We've reworked the "phone number search" logic a few times (moving to SSK, etc.).  Early on, I fixed a bug to make sure that we only show countries in the list with a valid name (non-empty) and a valid calling code (non-empty, not +0).  In all of the refactoring that fix got lost, so this fixes it again.

charlesmchen:
https://github.com/WhisperSystems/Signal-iOS/commit/00a29dea4d8d8bf6b9b8b8a68f97681fc97da157

PR #149, opened 2017-03-28
Accept arbitrary incoming attachments.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/OWSAttachmentsProcessor.m
+++ b/src/Messages/Attachments/OWSAttachmentsProcessor.m
@@ -79,7 +79,10 @@ NS_ASSUME_NONNULL_BEGIN
 
         [attachmentIds addObject:pointer.uniqueId];
 
-        if ([MIMETypeUtil isSupportedMIMEType:pointer.contentType]) {
+        // TODO: We need to revisit isSupportedMIMEType.
+        if ([MIMETypeUtil isSupportedMIMEType:pointer.contentType] ||
+            [pointer.contentType isEqualToString:OWSMimeTypeOversizeTextMessage] ||
+            [pointer.contentType isEqualToString:OWSMimeTypeUnknownForTests]) {
             [pointer save];
             [supportedAttachmentPointers addObject:pointer];
             [supportedAttachmentIds addObject:pointer.uniqueId];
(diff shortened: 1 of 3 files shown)


src/Messages/Attachments/OWSAttachmentsProcessor.m
```diff
@@ -79,7 +79,10 @@ - (instancetype)initWithAttachmentProtos:(NSArray<OWSSignalServiceProtosAttachme
 
         [attachmentIds addObject:pointer.uniqueId];
 
-        if ([MIMETypeUtil isSupportedMIMEType:pointer.contentType]) {
+        // TODO: We need to revisit isSupportedMIMEType.
+        if ([MIMETypeUtil isSupportedMIMEType:pointer.contentType] ||
+            [pointer.contentType isEqualToString:OWSMimeTypeOversizeTextMessage] ||
+            [pointer.contentType isEqualToString:OWSMimeTypeUnknownForTests]) {
```
charlesmchen, line 85: When we want to allow users to send arbitrary outgoing attachments as well, we should rework `isSupportedMIMEType`.

src/Util/MIMETypeUtil.h
```diff
@@ -52,4 +53,8 @@ extern NSString *const OWSMimeTypeOversizeTextMessage;
 + (NSSet<NSString *> *)supportedImageUTITypes;
 + (NSSet<NSString *> *)supportedAnimatedImageUTITypes;
 
++ (NSString *)utiTypeForMIMEType:(NSString *)mimeType;
++ (NSString *)fileExtensionForUTIType:(NSString *)utiType;
++ (NSString *)fileExtensionForMIMEType:(NSString *)mimeType;
```
charlesmchen, line 58: These functions are used in the new "arbitrary attachments" changes to Signal-iOS.

src/Util/MIMETypeUtil.m
```diff
@@ -13,6 +13,7 @@
 NSString *const OWSMimeTypeApplicationOctetStream = @"application/octet-stream";
 NSString *const OWSMimeTypeImagePng = @"image/png";
 NSString *const OWSMimeTypeOversizeTextMessage = @"text/x-signal-plain";
+NSString *const OWSMimeTypeUnknownForTests = @"unknown/mimetype";
```
michaelkirk, line 16: How are you using this for testing?

michaelkirk (approved):
LGTM

src/Util/MIMETypeUtil.m
charlesmchen: In the Debug UI, there will soon be an action to send an attachment with this MIME type, which should always be "unknown" to the recipient.

PR #154, opened 2017-03-30
Fix “two singletons” anti-pattern in TSMessagesManager.

charlesmchen:
We apparently were creating two instances of `TSMessagesManager`. One was created using the `sharedManager` singleton accessor pattern and one was created by `PushManager`.  

I also notice that we're creating multiple instances of `OWSMessageSender`.  I'm going to address that in a separate PR.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSMessagesManager.m
+++ b/src/Messages/TSMessagesManager.m
@@ -40,6 +40,36 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+static TSMessagesManager *shareMessagesManager = nil;
+
+id shareMessagesManagerLock() {
+    static id shareMessagesManagerLock = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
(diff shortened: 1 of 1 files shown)


src/Messages/TSMessagesManager.m
```diff
@@ -40,6 +40,36 @@
...
+}
+
+void SetShareMessagesManager(TSMessagesManager *value) {
+    @synchronized (shareMessagesManagerLock()) {
+        OWSCAssert(!shareMessagesManager);
+        OWSCAssert(value);
+        
+        shareMessagesManager = value;
+    }
+}
```
charlesmchen, line 69: Not the most elegant solution, but it'll allow us to inject TSMessagesManager's dependencies in `PushManager`.

charlesmchen:
I guess I'm wondering if there are other "singletons" that we're making more than one of.

charlesmchen:
Closing in favor of  #155.

PR #155, opened 2017-03-30
enforce singleton usage for MessagesManager and MessageSender

michaelkirk:
An alternative approach to #154 

Corresponds to iOS PR: https://github.com/WhisperSystems/Signal-iOS/pull/1917#pullrequestreview-30051758

PTAL @charlesmchen

### DIFF
--- a/src/Messages/TSMessagesManager.h
+++ b/src/Messages/TSMessagesManager.h
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 11/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSIncomingMessage.h"
 #import "TSInvalidIdentityKeySendingErrorMessage.h"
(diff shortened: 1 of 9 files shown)


src/Messages/TSMessagesManager.h
```diff
@@ -20,14 +21,6 @@ NS_ASSUME_NONNULL_BEGIN
 @interface TSMessagesManager : NSObject
 
 - (instancetype)init NS_UNAVAILABLE;
-
-- (instancetype)initWithNetworkManager:(TSNetworkManager *)networkManager
```
charlesmchen, line 24: If we're going to leave these constructors for tests, perhaps we can wrap them in `#ifdef DEBUG` so that we don't accidentally use them in production code.

charlesmchen (approved):
LGTM

src/Messages/TSMessagesManager.h
michaelkirk: I'm not sure I understand precisely what you are proposing.

We can't omit the definition from production.

Though we're no longer using this declaration in production, we still use the definition in production, it's just private, and ultimately used to implement `+ (instancetype)sharedManager`.

charlesmchen: Yup, I'm just proposing changes to this file - the header.

michaelkirk: You are proposing that instead of having the declaration only in the test file, put them in this file, but wrap them in `#ifdef debug`?

charlesmchen: That was the idea, but I'm open to alternatives.

michaelkirk: To me that doesn't seem any safer than *not* including the headers outside of tests at all, so I'm gong to leave it as is for now.

PR #156, opened 2017-03-31
Charlesmchen/blocking1

charlesmchen:
This PR builds upon https://github.com/WhisperSystems/SignalServiceKit/pull/155 which is not yet merged.

* Sketches out TSBlockingManager.
   * `TSBlockingManager` manages the block list.
   * `TSBlockingManager` persists the block list.
   * `TSBlockingManager` debounces redundant changes.
   * `TSBlockingManager` fires a `NSNotification` when the block list is changed.
   * `TSBlockingManager` fires "blocked phone number" sync messages when the block list is changed _locally_.
   * `TSMessagesManager` updates `TSBlockingManager` when a "blocked phone number" sync message is received from another client.
* `TSBlockingManager` tracks the last block list state that it was able to sync.  If block list syncing fails, the app will try again the next time it is launched. 
   * We could also retry.
* `TSBlockingManager` is intended to be as thread-safe as possible.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSBlockingManager.m
+++ b/src/Messages/TSBlockingManager.m
@@ -132,6 +132,18 @@ NSString *const kTSStorageManager_SyncedBlockedPhoneNumbersKey = @"kTSStorageMan
         if (!skipSyncMessage) {
             [self sendBlockedPhoneNumbersMessage:blockedPhoneNumbers];
         } else {
+            // If this update came from an incoming blocklist sync message,
+            // update the "synced blocked phone numbers" state immediately,
+            // since we're now in sync.
+            //
+            // There could be data loss if both clients modify the block list
+            // at the same time, but:
+            //
+            // a) Block list changes will be rare.
+            // b) Conflicting block list changes will be even rarer.
+            // c) It's unlikely a user will make conflicting changes on two
+            //    devices around the same time.
+            // d) There isn't a good way to avoid this.
             [self saveSyncedBlockedPhoneNumbers:blockedPhoneNumbers];
         }
 
@@ -186,6 +198,8 @@ NSString *const kTSStorageManager_SyncedBlockedPhoneNumbersKey = @"kTSStorageMan
         }
         failure:^(NSError *error) {
             DDLogError(@"%@ Failed to send blocked phone numbers with error: %@", self.tag, error);
+
+            // TODO: We might want to retry more often than just app launch.
         }];
 }
 
--- a/src/Messages/TSBlockingManager.m
+++ b/src/Messages/TSBlockingManager.m
@@ -1,0 +1,202 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "TSBlockingManager.h"
+#import "OWSBlockedPhoneNumbersMessage.h"
+#import "OWSMessageSender.h"
+#import "TSStorageManager.h"
+#import "TextSecureKitEnv.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSString * const kNSNotificationName_BlockedPhoneNumbersDidChange = @"kNSNotificationName_BlockedPhoneNumbersDidChange";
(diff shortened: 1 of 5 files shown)


src/Messages/TSBlockingManager.m
```diff
@@ -0,0 +1,239 @@
...
+        return self;
+    }
+
+    OWSAssert(storageManager);
+    OWSAssert(messageSender);
+    
+    _storageManager = storageManager;
+    _messageSender = messageSender;
+
+    [self loadBlockedPhoneNumbers];
```
michaelkirk: This is going to do blocking IO on initialization, which can't be in our startup path, lest a long running migration blocks the app from launching.

My first thought would be to move `loadBlockedPhoneNumbers` to a lazy accessor once we try to use the block list, which will require some other changes e.g. it would have to be synchronized.

Or maybe there's another way to ensure this initialization isn't called on our startup path?

src/Messages/TSBlockingManager.m
```diff
@@ -0,0 +1,239 @@
...
+// phone numbers.
+@property (nonatomic, readonly) NSMutableSet<NSString *> *blockedPhoneNumberSet;
+
+@end
+
+#pragma mark -
+
+@implementation TSBlockingManager
+
++ (instancetype)sharedManager {
```
michaelkirk: nit: clang-format

src/Messages/TSBlockingManager.m
```diff
@@ -0,0 +1,239 @@
...
+
+    @synchronized (self) {
+        if (![_blockedPhoneNumberSet containsObject:phoneNumber]) {
+            return;
+        }
+        
+        [_blockedPhoneNumberSet removeObject:phoneNumber];
+    }
+
+    [self handleUpdate:NO];
```
michaelkirk: nit: From the callers perspective, this intuitively reads as "do not handle the update"

maybe: `[self handleUpdateSkippingSyncMessage:NO]` 
or maybe even clearer to get rid of the double negation: `[self handleUpdateSendingSyncMessage:YES]`

src/Messages/TSBlockingManager.m
```diff
@@ -0,0 +1,239 @@
...
+}
+
+- (void)removeBlockedPhoneNumber:(NSString *)phoneNumber {
+    OWSAssert(phoneNumber.length > 0);
+
+    DDLogInfo(@"%@ removeBlockedPhoneNumber: %@", self.tag, phoneNumber);
+
+    @synchronized (self) {
+        if (![_blockedPhoneNumberSet containsObject:phoneNumber]) {
```
michaelkirk: nit: At first I wondered why you bothered with this "redundant" set logic, but I realized the goal of this early return is to avoid the call to `handleUpdate`. Maybe worth a comment here an in `addBlockedPhoneNumber`

michaelkirk (changes requested):

michaelkirk:
Overall approach looks good, just had a concern about reading from the DB on initialization since that might entail waiting for a while.

src/Messages/TSBlockingManager.m
charlesmchen: Done. Oddly clang-format wouldn't fix this.

src/Messages/TSBlockingManager.m
charlesmchen: Done.

src/Messages/TSBlockingManager.m
charlesmchen: Done.

src/Messages/TSBlockingManager.m
charlesmchen: I made a default flavor. 

One way people like to fix these readability problems is with using "boolean enums" so you can write code like `[self handleUpdate:SendSyncMessage::Yes]`.  It's definitely more readable but it's also tedious to write.

charlesmchen:
PTAL @michaelkirk

PR #157, opened 2017-04-01
Charlesmchen/singleton assert

charlesmchen:
There are probably more singletons, but these are the obvious candidates.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/TSAttachmentPointer.h
+++ b/src/Messages/Attachments/TSAttachmentPointer.h
@@ -17,11 +17,13 @@ typedef NS_ENUM(NSUInteger, TSAttachmentPointerState) {
  */
 @interface TSAttachmentPointer : TSAttachment
 
+- (instancetype)init NS_UNAVAILABLE;
+
 - (instancetype)initWithServerId:(UInt64)serverId
                              key:(NSData *)key
                           digest:(nullable NSData *)digest
                      contentType:(NSString *)contentType
-                           relay:(NSString *)relay NS_DESIGNATED_INITIALIZER;
+                           relay:(NSString *)relay;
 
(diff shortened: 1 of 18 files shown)


src/Messages/Attachments/TSAttachmentPointer.h
```diff
@@ -17,11 +17,13 @@ typedef NS_ENUM(NSUInteger, TSAttachmentPointerState) {
  */
 @interface TSAttachmentPointer : TSAttachment
 
+- (instancetype)init NS_UNAVAILABLE;
+
 - (instancetype)initWithServerId:(UInt64)serverId
                              key:(NSData *)key
                           digest:(nullable NSData *)digest
                      contentType:(NSString *)contentType
-                           relay:(NSString *)relay NS_DESIGNATED_INITIALIZER;
```
charlesmchen: This was causing two warnings.  I might be wrong, but I think the init + NS_UNAVAILABLE is better?

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1924

src/Messages/Attachments/TSAttachmentPointer.h
michaelkirk: Thanks for addressing our compiler warnings, but could you revert this change? 

It's not pertinent to the rest of this PR, and I'd like to address it differently - the changes I'd like to make will be more concise as a PR rather than a verbose comment here.

michaelkirk (approved):
LGTM. Thanks for fixing this.

src/Messages/Attachments/TSAttachmentPointer.h
charlesmchen: Done.

PR #158, opened 2017-04-01
Filter incoming and outgoing messages using the blacklist.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -17,6 +17,7 @@
 #import "SignalRecipient.h"
 #import "TSAccountManager.h"
 #import "TSAttachmentStream.h"
+#import "TSBlockingManager.h"
 #import "TSContactThread.h"
 #import "TSGroupThread.h"
 #import "TSIncomingMessage.h"
@@ -257,6 +258,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 
 @property (nonatomic, readonly) TSNetworkManager *networkManager;
 @property (nonatomic, readonly) TSStorageManager *storageManager;
+@property (nonatomic, readonly) TSBlockingManager *blockingManager;
 @property (nonatomic, readonly) OWSUploadingService *uploadingService;
 @property (nonatomic, readonly) YapDatabaseConnection *dbConnection;
 @property (nonatomic, readonly) id<ContactsManagerProtocol> contactsManager;
@@ -291,6 +293,14 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
     OWSSingletonAssert();
 
     return self;
+}
+
+- (void)setBlockingManager:(TSBlockingManager *)blockingManager
+{
+    OWSAssert(blockingManager);
+    OWSAssert(!_blockingManager);
+
+    _blockingManager = blockingManager;
 }
 
 - (NSOperationQueue *)sendingQueueForMessage:(TSOutgoingMessage *)message
@@ -516,10 +526,12 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 
             if (recipients.count == 0) {
                 if (error) {
-                    return failureHandler(error);
+                    failureHandler(error);
+                    return;
                 } else {
                     DDLogError(@"%@ Unknown error finding contacts", self.tag);
-                    return failureHandler(OWSErrorMakeFailedToSendOutgoingMessageError());
+                    failureHandler(OWSErrorMakeFailedToSendOutgoingMessageError());
+                    return;
                 }
             }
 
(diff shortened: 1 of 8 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -17,6 +17,7 @@
 #import "SignalRecipient.h"
 #import "TSAccountManager.h"
 #import "TSAttachmentStream.h"
+#import "TSBlockingManager.h"
```
michaelkirk: nit: I prefer the 3-letter class prefix (OWS) in line with apples style guide, to avoid the (admittedly small chance) that we're one day clobbered by a new apple framework.

src/Messages/OWSMessageSender.m
```diff
@@ -607,11 +630,31 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
     NSMutableArray<TOCFuture *> *futures = [NSMutableArray array];
 
+    NSArray<NSString *> *blockedPhoneNumbers = _blockingManager.blockedPhoneNumbers;
+    int blockedCount = 0;
     for (SignalRecipient *rec in recipients) {
-        // we don't need to send the message to ourselves, but otherwise we send
-        if (![[rec uniqueId] isEqualToString:[TSStorageManager localNumber]]) {
-            [futures addObject:[self sendMessageFuture:message recipient:rec thread:thread]];
+        // We don't need to send the message to ourselves...
+        if ([[rec uniqueId] isEqualToString:[TSStorageManager localNumber]]) {
```
michaelkirk: style nit: prefer `object.propertyName` for getters/setters, e.g. `rec.uniqueId`

src/Messages/OWSMessageSender.m
```diff
@@ -607,11 +630,31 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
+            continue;
         }
+        // ...or to anyone on our blocklist...
+        OWSAssert([rec uniqueId].length > 0);
+        if ([blockedPhoneNumbers containsObject:[rec uniqueId]]) {
```
michaelkirk: nit: maybe this would be better encapsulated as a method on the blockingManager?

`[blockingManager isBlockedRecipientIdentifier:rec.uniqueId]`

src/Messages/TSBlockingManager.m
```diff
@@ -66,6 +66,10 @@ - (instancetype)initStorageManager:(TSStorageManager *)storageManager messageSen
     _storageManager = storageManager;
```
michaelkirk, line 65: nit: I guess I missed this in the last CR, but shouldn't this method be called init**With**StorageManager, since it doesn't initialize **the** storage manager?

src/Messages/TSMessagesManager.m
```diff
@@ -155,6 +155,15 @@ - (void)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
     OWSAssert([NSThread isMainThread]);
 
     DDLogInfo(@"%@ received envelope: %@", self.tag, [self descriptionForEnvelope:envelope]);
+
+    // TODO: Can we trust envelope.source to be properly formatted?
```
michaelkirk: re: TODO 

Per previous discussion, this is a server provided value. As long as we continue to  use the same `source` to fetch the session identifier, any spoofed value here would not be decryptable.

We'd have to revisit this upon allowing non-phone identifiers.

michaelkirk (commented):
LGTM, just want to make sure we're handling blocking the same as Android/Desktop WRT groups.

src/Messages/OWSMessageSender.m
charlesmchen: Done.

src/Messages/OWSMessageSender.m
charlesmchen: Done. 

On a separate note, I've been thinking that we should add convenience accessors for some of these model objects.  In this case, I'd like to add `[SignalRecipient phoneNumberE164]` or perhaps `[SignalRecipient contactId]`.  We can discuss what's the best name, but the point is would be to distinguish whether or not we're looking for the database uniqueId or the phone number/signal account id.

src/Messages/OWSMessageSender.m
charlesmchen: I've been tempted myself.  The problem is that `OWSBlockingManager` does synchronization.  Some of the likely users of `[OWSBlockingManager isPhoneNumberBlocked:]` will be perf-sensitive, so I was wary of adding a method that we'd later use not realizing that it was a little expensive.  But yeah, it'd be nice to DRY this up.  I could go either way.

src/Messages/TSBlockingManager.m
charlesmchen: Done.

src/Messages/TSMessagesManager.m
charlesmchen: Done.

charlesmchen:
> LGTM, just want to make sure we're handling blocking the same as Android/Desktop WRT groups.

@michaelkirk https://github.com/WhisperSystems/SignalServiceKit/pull/159

PR #160, opened 2017-04-03
Retry block sync messages.

charlesmchen:
PTAL @michaelkirk


michaelkirk (commented):
There's a retry loop here, which calls the SendOperation which has it's own retry loop, which ultimately calls another retry loop when sending to the messages API endpoint. 

So I'm getting a little nervous about all this nested looping ending in a situation where we inadvertently rate-limit (or DDOS ourselves).

What does this get us that our MessageSender operation retry doesn't already get us?

PR #161, opened 2017-04-04
Don't retry some message sending failures

michaelkirk:
Motivation:

When we (I 😬) introduced the MessageSendingOperation, we included a new
"retry" loop. However, this had some unintended consequences when
retrying terminal failures.

Some of these are pretty benign and invisible to the user, but some,
like messaging someone who's safety number has changed, results in a
situation where we get rate-limited by the pre-key request.

Description:

This commit includes the machinery to distinguish between retryable and
terminal failures. Upon reporting a terminal failure, the MessageSender
stops retrying to send.

I've defaulted to retrying when at all reasonable, and give reasons whenever we *skip* the retry. Some of the reasons given are arguable. Feel free to argue.

PTAL @charlesmchen

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -79,7 +79,7 @@ typedef NS_ENUM(NSInteger, OWSSendMessageOperationState) {
 
 - (void)attemptToSendMessage:(TSOutgoingMessage *)message
                      success:(void (^)())successHandler
-                     failure:(void (^)(NSError *error))failureHandler;
+                     failure:(RetryableFailureHandler)failureHandler;
 
 @end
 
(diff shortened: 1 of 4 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -471,8 +480,15 @@ - (void)resendMessageFromKeyError:(TSInvalidIdentityKeySendingErrorMessage *)err
...
         // Normally marking as unsent is handled in sendMessage happy path, but beacuse we're skipping the common entry
         // point to message sending in order to send to a single recipient, we have to handle it ourselves.
-        void (^markAndFailureHandler)(NSError *error) = ^(NSError *error) {
+        RetryableFailureHandler markAndFailureHandler = ^(NSError *error, BOOL isRetryable) {
             [self saveMessage:message withError:error];
+            if (isRetryable) {
+                DDLogWarn(@"Skipping retry for message failure in this corner case.");
+                // Fixing this will require a larger refactor. In the meanwhile, we don't
+                // retry message sending from accepting key-changes in a group. (Except for the inner retry logic w/ the
+                // messages API)
```
charlesmchen, line 489: So at the moment we're only logging, but not changing the retry behavior here?

I tend to mark comments like this with a standard prefix: TODO: for readability and to make them grep-able later.

src/Messages/OWSMessageSender.m
```diff
@@ -616,7 +643,8 @@ - (TOCFuture *)sendMessageFuture:(TSOutgoingMessage *)message
         success:^{
             [futureSource trySetResult:@1];
         }
-        failure:^(NSError *error) {
+        failure:^(NSError *error, BOOL isRetryable) {
+            // FIXME what to do WRT retryable here?
```
charlesmchen, line 647: I see: you use FIXME.

charlesmchen (approved):
LGTM

michaelkirk:
in https://github.com/WhisperSystems/SignalServiceKit/commit/694088ee9829c7cec2dc78806b1e531152c05219

PR #162, opened 2017-04-06
ensure all session access happens on the sessionStoreQueue

michaelkirk:
Though mutating was already serialized via the encrypt/decrypt methods in SPK, there were some places where we were deleting and reading outside of the sessionStoreQueue

Doing so required adding a handful of dispatch_async's

PTAL @charlesmchen

### DIFF
--- a/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
+++ b/src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
@@ -1,5 +1,6 @@
-//  Created by Frederic Jacobs on 31/12/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 #import "TSInvalidIdentityKeyReceivingErrorMessage.h"
 #import "OWSFingerprint.h"
@@ -78,20 +79,25 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
-    [[TSStorageManager sharedManager] saveRemoteIdentity:newKey recipientId:self.envelope.source];
+    // Saving a new identity mutates the session store so it must happen on the sessionStoreQueue
+    dispatch_async([OWSDispatch sessionStoreQueue], ^{
+        [[TSStorageManager sharedManager] saveRemoteIdentity:newKey recipientId:self.envelope.source];
 
-    // Decrypt this and any old messages for the newly accepted key
-    NSArray<TSInvalidIdentityKeyReceivingErrorMessage *> *messagesToDecrypt =
-        [self.thread receivedMessagesForInvalidKey:newKey];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            // Decrypt this and any old messages for the newly accepted key
+            NSArray<TSInvalidIdentityKeyReceivingErrorMessage *> *messagesToDecrypt =
+                [self.thread receivedMessagesForInvalidKey:newKey];
 
-    for (TSInvalidIdentityKeyReceivingErrorMessage *errorMessage in messagesToDecrypt) {
-        [[TSMessagesManager sharedManager] handleReceivedEnvelope:errorMessage.envelope];
+            for (TSInvalidIdentityKeyReceivingErrorMessage *errorMessage in messagesToDecrypt) {
+                [[TSMessagesManager sharedManager] handleReceivedEnvelope:errorMessage.envelope];
 
-        // Here we remove the existing error message because handleReceivedEnvelope will either
-        //  1.) succeed and create a new successful message in the thread or...
-        //  2.) fail and create a new identical error message in the thread.
-        [errorMessage remove];
-    }
+                // Here we remove the existing error message because handleReceivedEnvelope will either
+                //  1.) succeed and create a new successful message in the thread or...
+                //  2.) fail and create a new identical error message in the thread.
+                [errorMessage remove];
+            }
+        });
+    });
 }
 
 - (NSData *)newIdentityKey
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -42,6 +42,13 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+void AssertIsOnSendingQueue()
+{
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(10, 0)) {
+        dispatch_assert_queue([OWSDispatch sendingQueue]);
+    } // else, skip assert as it's a development convenience.
+}
+
 /**
  * OWSSendMessageOperation encapsulates all the work associated with sending a message, e.g. uploading attachments,
  * getting proper keys, and retrying upon failure.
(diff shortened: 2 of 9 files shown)


src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
```diff
@@ -78,20 +79,25 @@ - (void)acceptNewIdentityKey
...
+            for (TSInvalidIdentityKeyReceivingErrorMessage *errorMessage in messagesToDecrypt) {
+                [[TSMessagesManager sharedManager] handleReceivedEnvelope:errorMessage.envelope];
+
+                // Here we remove the existing error message because handleReceivedEnvelope will either
+                //  1.) succeed and create a new successful message in the thread or...
+                //  2.) fail and create a new identical error message in the thread.
+                [errorMessage remove];
+            }
+        });
+    });
```
charlesmchen, line 100: It looks like `acceptNewIdentityKey` is only being called from `MessagesViewController`.  Immediately after, `MessagesViewController` tries to resend a message.  Now that `acceptNewIdentityKey`, should it take a completion block so that we can ensure that the resend occurs after the acceptance is complete?

src/Messages/OWSMessageSender.m
```diff
@@ -42,10 +42,12 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
-#define SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(major, minor)                                                          \
-    ([[NSProcessInfo processInfo]                                                                                      \
-        isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){                                                    \
-                                            .majorVersion = major, .minorVersion = minor, .patchVersion = 0 }])
+void AssertIsOnSendingQueue()
+{
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(10, 0)) {
+        dispatch_assert_queue([OWSDispatch sendingQueue]);
```
charlesmchen, line 48: It's not always safe to assert the current queue. Sometimes when you dispatch_sync onto another queue you don't actually switch queues as an optimization. 

From https://developer.apple.com/reference/dispatch/1452870-dispatch_sync

> As an optimization, this function invokes the block on the current thread when possible.

So this is only safe if we're using dispatch_async(), etc.  And I tend to only assert that we're on the main queue since a) I strenuously avoid doing dispatch_sync() onto the main queue. b) even if this is safe today, tomorrow someone might dispatch_sync() to code with this assertion.

src/Messages/OWSMessageSender.m
```diff
@@ -862,8 +865,7 @@ - (void)sendMessage:(TSOutgoingMessage *)message
                         return failureHandler(error, YES);
                     }
 
-                    [self handleMismatchedDevices:serializedResponse recipient:recipient];
-                    retrySend();
+                    [self handleMismatchedDevices:serializedResponse recipient:recipient completion:retrySend];
```
charlesmchen, line 868: Correct me if I'm wrong but won't these changes move invocation of `retrySend()` to the `sessionStoreQueue`?  Is that okay?  Should we modify `handleMismatchedDevices` and `handleStaleDevicesWithResponse` to execute their new completion handlers on another queue?

src/Messages/OWSMessageSender.m
charlesmchen: Upon reflection, I guess the benefits of this outweigh the risks. We'll never dispatch_sync to the sending queue, right?

src/Storage/AxolotlStore/TSStorageManager+SessionStore.m
```diff
@@ -6,12 +6,21 @@
 
 #define TSStorageManagerSessionStoreCollection @"TSStorageManagerSessionStoreCollection"
 
-@implementation TSStorageManager (SessionStore)
+void AssertIsOnSessionStoreQueue()
+{
+    if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(10, 0)) {
+        dispatch_assert_queue([OWSDispatch sessionStoreQueue]);
```
charlesmchen, line 12: Same discussion. We'll never `dispatch_sync()`, so this is fine, right?

src/Util/OWSDispatch.h
```diff
@@ -15,7 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
  * Signal protocol session state must be coordinated on a serial queue. This is sometimes used synchronously,
  * so never dispatching sync *from* this queue to avoid deadlock.
```
charlesmchen, line 16: I wonder if we should elaborate this comment to note that all encryption and decryption should take place on this queue?

src/Util/iOSVersions.h
```diff
@@ -0,0 +1,9 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#include <Availability.h>
+
+#define SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(major, minor) \
```
charlesmchen, line 8: Can we remove the definition of this in SessionCipher?

src/Messages/OWSMessageSender.m
michaelkirk: I understand the risks of dispatch_sync. But I don't understand your concern with dispatch_assert_queue.

What is the danger introduced by adding an *assert*? The *assert* doesn't do any dispatch - it just blows up when violated in DEBUG. Right?

src/Messages/OWSMessageSender.m
michaelkirk: retrySend already dispatches async to the send queue. Do you think that concern should be moved?

src/Messages/OWSMessageSender.m
charlesmchen: Well, a) it might lead to spurious asserts in development and b) I could be wrong but my impression was that dispatch_assert_queue() only applies in debug builds.

src/Util/iOSVersions.h
michaelkirk: Yeah, we *could* move iOSVersions to SignalProtocolKit. I guess I was being lazy until we used it there twice.

src/Messages/OWSMessageSender.m
charlesmchen: Ah fine!

src/Util/iOSVersions.h
charlesmchen: No biggee.

src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyReceivingErrorMessage.m
michaelkirk: Currently this is fine because any session `getting` will have to be dispatched to the sessionStoreQueue, which would be run serially *after* these changes. But it would be more resilient to refactor if we did that in a completion handler. I'll make the change if it's as straight forward as I hope.

src/Util/OWSDispatch.h
michaelkirk: It's the sessionState mutation that we actually care about though, right?

src/Messages/OWSMessageSender.m
michaelkirk: Also keep in mind I'm asserting *queues* not threads here.

michaelkirk: > it might lead to spurious asserts in development

My understanding is there will be no spurious asserts. 

> my impression was that dispatch_assert_queue() only applies in debug builds

That was my assumption, but actually turns out the failed asserts will crash. I've added debug checks at https://github.com/WhisperSystems/SignalServiceKit/pull/163 and https://github.com/WhisperSystems/SignalProtocolKit/pull/19

PR #163, opened 2017-04-07
only assert queues in debug

michaelkirk:
PTAL @charlesmchen


charlesmchen (approved):
LGTM thanks for following up.

PR #164, opened 2017-04-08
Charlesmchen/socket lifecycle

charlesmchen:
I've reworked the socket manager.

* Simplify socket manager's interface.
* Improve documentation throughout socket manager.
* Modify socket manager to observe application state and registration state events.
* Simplify, centralize and make explicit the logic that determines whether or not the socket manager "desires" the socket to be open.
  * Remove conflicts between the previously separate logic that was trying to keep the web socket open if the app was in the foreground, we had received a notification in the background, or we had received a message in the background.
* Simplify and unify state changes to ensure that manager properties are always reset/updated properly.
* Ensure that we only create "background tasks" when the app is inactive or in the background.

This fixes at least one serious bug and aims to reduce the surface area for bugs.  

This change is not for v2.10. We need to dogfood this extensively.

PTAL @michaelkirk

### DIFF
--- a/src/Account/TSAccountManager.m
+++ b/src/Account/TSAccountManager.m
@@ -15,12 +15,18 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSString *const TSRegistrationErrorDomain = @"TSRegistrationErrorDomain";
+NSString *const TSRegistrationErrorUserInfoHTTPStatus = @"TSHTTPStatus";
+NSString *const kNSNotificationName_RegistrationStateDidChange = @"kNSNotificationName_RegistrationStateDidChange";
+
 @interface TSAccountManager ()
 
 @property (nullable, nonatomic, retain) NSString *phoneNumberAwaitingVerification;
 @property (nonatomic, strong, readonly) TSStorageManager *storageManager;
 
 @end
+
+#pragma mark -
 
 @implementation TSAccountManager
 
@@ -71,6 +77,10 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     [self.storageManager storePhoneNumber:phoneNumber];
+
+    [[NSNotificationCenter defaultCenter] postNotificationName:kNSNotificationName_RegistrationStateDidChange
+                                                        object:nil
+                                                      userInfo:nil];
 }
 
 + (nullable NSString *)localNumber
@@ -204,7 +214,7 @@ NS_ASSUME_NONNULL_BEGIN
                 case 204: {
                     [TSStorageManager storeServerToken:authToken signalingKey:signalingKey];
                     [self didRegister];
-                    [TSSocketManager becomeActiveFromForeground];
+                    [TSSocketManager requestSocketOpen];
                     [TSPreKeyManager registerPreKeysWithMode:RefreshPreKeysMode_SignedAndOneTime
                                                      success:successBlock
                                                      failure:failureBlock];
@@ -260,6 +270,11 @@ NS_ASSUME_NONNULL_BEGIN
         success:^(NSURLSessionDataTask *task, id responseObject) {
             DDLogInfo(@"%@ Successfully unregistered", self.tag);
             success();
+
+            // The success handler should reset local storage.
+            [[NSNotificationCenter defaultCenter] postNotificationName:kNSNotificationName_RegistrationStateDidChange
+                                                                object:nil
+                                                              userInfo:nil];
         }
         failure:^(NSURLSessionDataTask *task, NSError *error) {
             DDLogError(@"%@ Failed to unregister with error: %@", self.tag, error);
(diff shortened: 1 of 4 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1953

charlesmchen:
Not for v2.10. Benefits are very real but too sensitive to introduce this late.

src/Account/TSAccountManager.m
```diff
@@ -260,6 +270,11 @@ + (void)unregisterTextSecureWithSuccess:(void (^)())success failure:(void (^)(NS
         success:^(NSURLSessionDataTask *task, id responseObject) {
             DDLogInfo(@"%@ Successfully unregistered", self.tag);
             success();
+
+            // The success handler should reset local storage.
```
michaelkirk: >  // The success handler should reset local storage.

Just to clarify, it seems like this is a TODO rather than a documentation of current behavior. Could you mark it with TODO?

The current code layout makes this seem related to the subsequent `NSNotificationCenter` line, but I think it's a separate concern, right? If so, could you insert a newline.

Also, I'm not sure that we want to reset storage here. E.g. Maybe we'd rather let people re-register, as you can on android, without losing your local history. Reason being to soften the blow a bit when someone accidentally activates a new device without realizing it de-auths their old device.

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -425,157 +461,234 @@ - (NSString *)webSocketAuthenticationString {
...
+#pragma mark - Socket LifeCycle
+
+- (BOOL)shouldSocketBeOpen
+{
     OWSAssert([NSThread isMainThread]);
-    
-    if (self.fetchingTaskIdentifier != UIBackgroundTaskInvalid) {
-        [self closeBackgroundTask];
+
+    if (![TSAccountManager isRegistered]) {
```
michaelkirk, line 520: We should avoid introducing any blocking DB i/o to our app launch startup path, since it can cause startup to hang. e.g. put a break point here and see how the app hangs on the launch screen.

I've seen DB reading result in a repeated crash-on-launch when a new DB extension is added. It's tricky because it bites you later. Specifically, adding a new db extension enumerates all objects, which can take a long time if the user has a large message database and is on an old device.

Similarly, it looks like we're also reading from the DB in `ensureWebsocketIsOpen` in order to get the websocket credentials. We previously got the websocket credentials in `ensureWebsocket` but that dispatched async, which maybe sidestepped the issue.

AFAICT this is only the result of the newly observed UIApplicationDidBecomeActiveNotification. Can we delay observing that notification until the app has successfully launched?

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -317,7 +355,10 @@ - (void)processWebSocketRequestMessage:(WebSocketRequestMessage *)message {
     DDLogInfo(@"Got message with verb: %@ and path: %@", message.verb, message.path);
 
     [self sendWebSocketMessageAcknowledgement:message];
-    [self keepAliveBackground];
+
+    // If we receive a message over the socket while the app is in the background,
+    // prolong how long the socket stays open.
+    [self requestSocketAliveForAtLeastSeconds:kBackgroundKeepSocketAliveDurationSeconds];
```
michaelkirk, line 375: 👍

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -425,157 +461,234 @@ - (NSString *)webSocketAuthenticationString {
                          [TSStorageManager serverAuthToken]];
 }
 
-- (BOOL)shouldKeepWebSocketAlive {
-    OWSAssert([NSThread isMainThread]);
+- (void)setFetchingTaskIdentifier:(UIBackgroundTaskIdentifier)fetchingTaskIdentifier
+{
```
michaelkirk: nit: seems like we could use the automatically synthesized setter?

michaelkirk (commented):
Just the one blocking concern re: startup. 

Nice work though, I'm eager to get this merged and get some miles on it.

src/Account/TSAccountManager.m
charlesmchen: > it seems like this is a TODO rather than a documentation of current behavior. 

Nope, this document current behavior.  This is called from `[SettingsTableViewController proceedToUnregistration]` whose success handler calls `[Environment resetAppData]`.   This method, after calling that success handler, then fires `kNSNotificationName_RegistrationStateDidChange` which is only safe to fire after the data store is reset.  This isn't a great arrangement, but I can see why we landed here (SSK doesn't know about `Environment`, etc.).  Hence the comment.  

I've elaborated the comment to more clear.

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Yup, thanks.

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Good point. I've updated this PR.

charlesmchen:
PTAL @michaelkirk

michaelkirk (approved):
LGTM

PR #165, opened 2017-04-10
Create block offer when non-contacts send you a message.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/Interactions/TSErrorMessage.m
+++ b/src/Messages/Interactions/TSErrorMessage.m
@@ -1,9 +1,5 @@
 //
-//  TSErrorMessage.m
-//  TextSecureKit
-//
-//  Created by Frederic Jacobs on 12/11/14.
-//  Copyright (c) 2014 Open Whisper Systems. All rights reserved.
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
 #import "TSErrorMessage.h"
@@ -36,8 +32,14 @@
     }
 
     _errorType = errorMessageType;
-    // TODO Move this out of model class.
-    [[TextSecureKitEnv sharedEnv].notificationsManager notifyUserForErrorMessage:self inThread:thread];
+    // TODO: Move this out of model class.
+    //
+    //       For now, dispatch async to ensure we're not inside a transaction
+    //       and thereby avoid deadlock.
+    TSErrorMessage *errorMessage = self;
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        [[TextSecureKitEnv sharedEnv].notificationsManager notifyUserForErrorMessage:errorMessage inThread:thread];
+    });
 
     return self;
 }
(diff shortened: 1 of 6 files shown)


src/Messages/Interactions/TSErrorMessage.m
```diff
@@ -36,8 +32,14 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
...
 
     _errorType = errorMessageType;
-    // TODO Move this out of model class.
-    [[TextSecureKitEnv sharedEnv].notificationsManager notifyUserForErrorMessage:self inThread:thread];
+    // TODO: Move this out of model class.
+    //
+    //       For now, dispatch async to ensure we're not inside a transaction
+    //       and thereby avoid deadlock.
+    TSErrorMessage *errorMessage = self;
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
charlesmchen, line 40: `notifyUserForErrorMessage: inThread :` eventually uses a transaction within `PropertyListPreferences`, which caused deadlock if you created any `TSErrorMessage` within a transaction.

src/Messages/OWSUnknownContactBlockOfferMessage.m
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return self;
+}
+
+- (nullable NSDate *)receiptDateForSorting
+{
+    // Always use date, since we're creating these interactions after the fact
+    // and back-dating them.
+    return self.date;
```
charlesmchen: This isn't the most elegant solution but I feel like its better than mutating `receivedAtDate`.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1960

charlesmchen:
Not for v2.10.

src/Messages/Interactions/TSErrorMessage.m
```diff
@@ -47,7 +49,7 @@ - (instancetype)initWithEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
                failedMessageType:(TSErrorMessageType)errorMessageType
 {
     TSContactThread *contactThread =
-        [TSContactThread getOrCreateThreadWithContactId:envelope.source transaction:transaction];
+        [TSContactThread getOrCreateThreadWithContactId:envelope.source transaction:transaction relay:nil];
```
michaelkirk: why this change?

src/Messages/OWSUnknownContactBlockOfferMessage.m
michaelkirk: There is some nuance to our different date fields, but the comment doesn't really clear it up. Could you elaborate on the behavior this code is intended to reproduce?

michaelkirk (approved):
LGTM, though see my coment in Signal-iOS re: using an InfoMessage rather than an ErrorMessage.

src/Messages/OWSUnknownContactBlockOfferMessage.m
charlesmchen: Done.

src/Messages/Interactions/TSErrorMessage.m
charlesmchen: Reverted.

PR #166, opened 2017-04-10
Remove the properties related to Redphone and WebRTC support.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
Not for v2.10.

michaelkirk (approved):
LGTM pending confirmation that other clients don't use these fields.

charlesmchen:
Per Moxie, we're not going to merge this until we tear down the RedPhone infrastructure.

michaelkirk:
>> LGTM pending confirmation that other clients don't use these fields.
> Per Moxie, we're not going to merge this until we tear down the RedPhone infrastructure.

You know what, that was my bad. This PR is just affecting what we're storing locally for other recipients - not what we're sending the accounts API, where we've already hardcoded those fields to be YES: https://github.com/WhisperSystems/SignalServiceKit/blob/master/src/Account/TSAttributes.m

I think we can merge this post haste. Right?

PR #167, opened 2017-04-10
Don’t mark messages as failed until all retries are exhausted.

charlesmchen:
This explains why attachment failures were leaving message sending in a bad state and subsequent messages were never sent.  We were marking message send failures as failed after the first failure, not when all retries were exhausted.  With attachments, you tend to notice this because they take so long to send.  

This isn't a regression, but now that we've raised the attachment file size limits, people will see this much more often (since the frequency of transfer failure varies as the square of the message size/message send duration).  

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -73,6 +73,8 @@ void AssertIsOnSendingQueue()
 
 @end
 
+#pragma mark -
+
 typedef NS_ENUM(NSInteger, OWSSendMessageOperationState) {
     OWSSendMessageOperationStateNew,
     OWSSendMessageOperationStateExecuting,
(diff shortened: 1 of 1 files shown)


charlesmchen:
v2.10 hotfix.

src/Messages/OWSMessageSender.m
```diff
@@ -228,6 +236,9 @@ - (void)tryWithRemainingRetries:(NSUInteger)remainingRetries
             [self tryWithRemainingRetries:remainingRetries - 1];
         } else {
             DDLogWarn(@"%@ Too many failures. Giving up sending.", self.tag);
+
+            [self.messageSender saveMessage:self.message withError:error];
```
michaelkirk: This also needs to happen for terminal failures (like SN changes). 

You could copy/paste it above, or probably better, move to the failureHandler wrapper in `initWithMessage `.

src/Messages/OWSMessageSender.m
```diff
@@ -363,17 +374,20 @@ - (void)attemptToSendMessage:(TSOutgoingMessage *)message
                      failure:(RetryableFailureHandler)failureHandler
 {
     DDLogDebug(@"%@ sending message: %@", self.tag, message.debugDescription);
-    RetryableFailureHandler markAndFailureHandler = ^(NSError *error, BOOL isRetryable) {
-        // TODO do we really want to mark this as failed if we're still retrying?
```
michaelkirk, line 367: death by TODO. I failed to recognize that this was pertinent. =(

michaelkirk (approved):
Except for the one change, LGTM.

src/Messages/OWSMessageSender.m
charlesmchen: Good catch.

PR #168, opened 2017-04-10
Improve attachment upload progress indicator.

charlesmchen:
* Should never show up for attachments in the failed state.
* Show show up without delay for attachments being sent.
* Show disappear when attachment upload fails in any way.

PTAL @michaelkirk 

For v2.11.


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1961

michaelkirk (approved):
LGTM

PR #170, opened 2017-04-10
Fix multiple match for SignalRecipient

### LINKED ISSUE #146: Possible bug in ContactsUpdater
I was looking through the code to understand how the protocol works, and I noticed something odd:

In the `ContactsUpdater`class, the `lookupIdentifiers` method will execute the `failure` handler when the set returned from `contactIntersectionWithSet` has an element count unequal to 1.

Here is the code in question:

````
[self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
                             success:^(NSSet<NSString *> *_Nonnull matchedIds) {
                                 if (matchedIds.count == 1) { // Shouldn't this be < 1
                                     NSMutableArray<SignalRecipient *> *recipients = [NSMutableArray new];
                                     for (NSString *identifier in matchedIds) {
                                         [recipients addObject:[SignalRecipient recipientWithTextSecureIdentifier:identifier]];
                                     }
                                     success([recipients copy]);
                                 } else {
                                     failure(OWSErrorMakeNoSuchSignalRecipientError());
                                 }
                             }
                             failure:failure];
````

The way I understand the code this should be
`if (matchedIds.count < 1)` instead of `if (matchedIds.count == 1)`.

Maybe it is intentional to only return one recipient, but then I don't understand why the for loop is placed there, and why a set is returned instead of a single object. And the `OWSErrorMakeNoSuchSignalRecipientError` would not really be correct, if the set contains more than 1 matching IDs

michaelkirk:
...when searching for contacts by phone number

In practice this was unlikely to be an issue, since it's unlikely you'd match more than one signal recipient since the array of identifiers are so similar.

fixes https://github.com/WhisperSystems/SignalServiceKit/issues/146

PTAL @charlesmchen

### DIFF
--- a/src/Contacts/ContactsUpdater.m
+++ b/src/Contacts/ContactsUpdater.m
@@ -94,10 +94,10 @@ NS_ASSUME_NONNULL_BEGIN
         failure(OWSErrorWithCodeDescription(OWSErrorCodeInvalidMethodParameters, @"Cannot lookup zero identifiers"));
         return;
     }
-    
+
     [self contactIntersectionWithSet:[NSSet setWithArray:identifiers]
                              success:^(NSSet<NSString *> *_Nonnull matchedIds) {
-                                 if (matchedIds.count == 1) {
+                                 if (matchedIds.count > 0) {
                                      NSMutableArray<SignalRecipient *> *recipients = [NSMutableArray new];
                                      for (NSString *identifier in matchedIds) {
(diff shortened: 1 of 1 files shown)


charlesmchen (approved):
LGTM I think I opened a Trello card too somewhere.

michaelkirk:
In https://github.com/WhisperSystems/SignalServiceKit/commit/f94021df9d690c9189f6cb5936e6d7af7705df06

PR #172, opened 2017-04-11
Fix file extensions for arbitrary file types.

charlesmchen:
PTAL @michaelkirk


src/Util/MIMETypeUtil.m
```diff
@@ -452,4 +453,1087 @@ + (NSString *)fileExtensionForMIMEType:(NSString *)mimeType
     return result;
 }
 
++ (NSDictionary *)genericMIMETypesToExtensionTypes
+{
+    static NSDictionary *result = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        result = @{
```
michaelkirk, line 461: Can you leave a comment about the source of authority for this list in case there are errors/updates in the future?

michaelkirk (approved):
LGTM.

src/Util/MIMETypeUtil.m
charlesmchen: Per offline conversation, this is a mishmash from many sources, manually collated.

PR #173, opened 2017-04-11
Rework outgoing message state.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/Interactions/TSOutgoingMessage.h
+++ b/src/Messages/Interactions/TSOutgoingMessage.h
@@ -34,6 +34,10 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 
 @interface TSOutgoingMessage : TSMessage
 
+- (instancetype)initWithTimestamp:(uint64_t)timestamp;
+
+- (instancetype)initWithTimestamp:(uint64_t)timestamp inThread:(nullable TSThread *)thread;
+
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(nullable TSThread *)thread
                       messageBody:(nullable NSString *)body;
@@ -41,6 +45,11 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(nullable TSThread *)thread
                  groupMetaMessage:(TSGroupMetaMessage)groupMetaMessage;
+
+- (instancetype)initWithTimestamp:(uint64_t)timestamp
+                         inThread:(nullable TSThread *)thread
+                      messageBody:(nullable NSString *)body
+                    attachmentIds:(NSMutableArray<NSString *> *)attachmentIds;
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(nullable TSThread *)thread
@@ -153,6 +162,7 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 
 #pragma mark - Sent Recipients
 
+- (NSUInteger)sentRecipientsCount;
 - (BOOL)wasSentToRecipient:(NSString *)contactId;
 - (void)updateWithSentRecipient:(NSString *)contactId transaction:(YapDatabaseReadWriteTransaction *)transaction;
 - (void)updateWithSentRecipient:(NSString *)contactId;
--- a/src/Messages/Interactions/TSOutgoingMessage.h
+++ b/src/Messages/Interactions/TSOutgoingMessage.h
@@ -77,6 +77,8 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
 // A map of attachment id-to-filename.
 @property (nonatomic, readonly) NSMutableDictionary<NSString *, NSString *> *attachmentFilenameMap;
 
+@property (atomic, readonly) TSGroupMetaMessage groupMetaMessage;
+
 /**
  * Whether the message should be serialized as a modern aka Content, or the old style legacy message.
  * Sync and Call messsages must be sent as Content, but other old style DataMessage payloads should be
--- a/src/Messages/Interactions/TSOutgoingMessage.h
+++ b/src/Messages/Interactions/TSOutgoingMessage.h
@@ -109,18 +109,30 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
                                                                         filename:(nullable NSString *)filename;
 
 // TSOutgoingMessage are updated from many threads. We don't want to save
-// our local copy since it may be out of date.  Instead, we use these
-// "updateWith..." methods to:
+// our local copy (this instance) since it may be out of date.  Instead, we
+// use these "updateWith..." methods to:
 //
-// a) Update a property of the local copy.
+// a) Update a property of this instance.
 // b) Load an up-to-date instance of this model from from the data store.
 // c) Update and save that fresh instance.
+// d) If this message hasn't yet been saved, save this local instance.
+//
+// After "updateWith...":
+//
+// a) An updated copy of this message will always have been saved in the
+//    data store.
+// b) The local property on this instance will always have been updated.
+// c) Other properties on this instance may be out of date.
+//
+// All mutable properties of this class have been made read-only to
+// prevent accidentally modifying them directly.
 //
 // This isn't a perfect arrangement, but in practice this will prevent
-// data loss.
+// data loss and will resolve all known issues.
 - (void)updateWithMessageState:(TSOutgoingMessageState)messageState;
 - (void)updateWithSendingError:(NSError *)error;
 - (void)updateWithHasSyncedTranscript:(BOOL)hasSyncedTranscript;
+- (void)updateWithCustomMessage:(NSString *)customMessage transaction:(YapDatabaseReadWriteTransaction *)transaction;
 - (void)updateWithCustomMessage:(NSString *)customMessage;
 - (void)updateWithWasDeliveredWithTransaction:(YapDatabaseReadWriteTransaction *)transaction;
 - (void)updateWithWasDelivered;
--- a/src/Messages/Interactions/TSOutgoingMessage.h
+++ b/src/Messages/Interactions/TSOutgoingMessage.h
@@ -19,11 +19,11 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
     TSOutgoingMessageStateAttemptingOut,
     // The failure state.
     TSOutgoingMessageStateUnsent,
-    // The message has been sent to the service, but not received by any recipient client.
-    TSOutgoingMessageStateSent,
-    // The message has been sent to the service and received by at least one recipient client.
-    // A recipient may have more than one client, and group message may have more than one recipient.
-    TSOutgoingMessageStateDelivered
+    // These two enum values have been combined into TSOutgoingMessageStateSentToService.
+    TSOutgoingMessageStateSent_OBSOLETE,
+    TSOutgoingMessageStateDelivered_OBSOLETE,
+    // The message has been sent to the service.
+    TSOutgoingMessageStateSentToService,
 };
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp;
@@ -53,9 +53,14 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
 
 - (instancetype)initWithCoder:(NSCoder *)coder NS_DESIGNATED_INITIALIZER;
 
-@property (nonatomic) TSOutgoingMessageState messageState;
-@property BOOL hasSyncedTranscript;
-@property NSString *customMessage;
+@property (atomic, readonly) TSOutgoingMessageState messageState;
+
+// The message has been sent to the service and received by at least one recipient client.
+// A recipient may have more than one client, and group message may have more than one recipient.
+@property (atomic, readonly) BOOL wasDelivered;
+
+@property (atomic, readonly) BOOL hasSyncedTranscript;
+@property (atomic, readonly) NSString *customMessage;
 @property (atomic, readonly) NSString *mostRecentFailureText;
 // A map of attachment id-to-filename.
 @property (nonatomic, readonly) NSMutableDictionary<NSString *, NSString *> *attachmentFilenameMap;
@@ -66,8 +71,6 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
  * sent as legacy message until we're confident no significant number of legacy clients exist in the wild.
  */
 @property (nonatomic, readonly) BOOL isLegacyMessage;
-
-- (void)setSendingError:(NSError *)error;
 
 /**
  * Signal Identifier (e.g. e164 number) or nil if in a group thread.
@@ -104,6 +107,29 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
  */
 - (OWSSignalServiceProtosAttachmentPointer *)buildAttachmentProtoForAttachmentId:(NSString *)attachmentId
                                                                         filename:(nullable NSString *)filename;
+
+// TSOutgoingMessage are updated from many threads. We don't want to save
+// our local copy since it may be out of date.  Instead, we use these
+// "updateWith..." methods to:
+//
+// a) Update a property of the local copy.
+// b) Load an up-to-date instance of this model from from the data store.
+// c) Update and save that fresh instance.
+//
+// This isn't a perfect arrangement, but in practice this will prevent
+// data loss.
+- (void)updateWithMessageState:(TSOutgoingMessageState)messageState;
+- (void)updateWithSendingError:(NSError *)error;
+- (void)updateWithHasSyncedTranscript:(BOOL)hasSyncedTranscript;
+- (void)updateWithCustomMessage:(NSString *)customMessage;
+- (void)updateWithWasDeliveredWithTransaction:(YapDatabaseReadWriteTransaction *)transaction;
+- (void)updateWithWasDelivered;
+
+#pragma mark - Sent Recipients
+
+- (BOOL)wasSentToRecipient:(NSString *)contactId;
+- (void)updateWithSentRecipient:(NSString *)contactId transaction:(YapDatabaseReadWriteTransaction *)transaction;
+- (void)updateWithSentRecipient:(NSString *)contactId;
 
 @end
 
--- a/src/Messages/Interactions/TSOutgoingMessage.h
+++ b/src/Messages/Interactions/TSOutgoingMessage.h
@@ -6,11 +6,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class OWSSignalServiceProtosAttachmentPointer;
-@class OWSSignalServiceProtosDataMessageBuilder;
-
-@interface TSOutgoingMessage : TSMessage
-
 typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
     // The message is either:
     // a) Enqueued for sending.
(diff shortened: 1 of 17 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1974

src/Messages/Interactions/TSOutgoingMessage.h
```diff
@@ -53,9 +53,14 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
...
-@property NSString *customMessage;
+@property (atomic, readonly) TSOutgoingMessageState messageState;
+
+// The message has been sent to the service and received by at least one recipient client.
+// A recipient may have more than one client, and group message may have more than one recipient.
+@property (atomic, readonly) BOOL wasDelivered;
+
+@property (atomic, readonly) BOOL hasSyncedTranscript;
+@property (atomic, readonly) NSString *customMessage;
 @property (atomic, readonly) NSString *mostRecentFailureText;
```
charlesmchen, line 85: These properties are used from multiple threads so they should all be atomic.

src/Messages/Interactions/TSOutgoingMessage.h
```diff
@@ -99,6 +102,29 @@ typedef NS_ENUM(NSInteger, TSOutgoingMessageState) {
...
+// TSOutgoingMessage are updated from many threads. We don't want to save
+// our local copy since it may be out of date.  Instead, we use these
+// "updateWith..." methods to:
+//
+// a) Update a property of the local copy.
+// b) Load an up-to-date instance of this model from from the data store.
+// c) Update and save that fresh instance.
+//
+// This isn't a perfect arrangement, but in practice this will prevent
+// data loss.
```
charlesmchen: This is the key change.

src/Messages/OWSMessageSender.m
```diff
@@ -486,36 +487,7 @@ - (void)resendMessageFromKeyError:(TSInvalidIdentityKeySendingErrorMessage *)err
     //  2.) fail and create a new identical error message in the thread.
     [errorMessage remove];
 
-    if ([errorMessage.thread isKindOfClass:[TSContactThread class]]) {
```
charlesmchen, line 494: This is a significant change. When re-sending group messages that failed to safety number changes, we used a special code path to avoid double-sending to other members of the group.  That code path was mostly redundant with our normal message sending code path but had some differences. Most significantly it ignored the message send operation queue.  I've consolidated along the one true code path, for consistency and correctness.  We now avoid double-sending to other members of a group using another mechanism which also applies to message send retries.

src/Messages/OWSMessageSender.m
```diff
@@ -659,6 +631,8 @@ - (TOCFuture *)sendMessageFuture:(TSOutgoingMessage *)message
         thread:thread
         attempts:OWSMessageSenderRetryAttempts
         success:^{
+            DDLogInfo(@"Marking group message as sent to recipient: %@", recipient.uniqueId);
+            [message updateWithSentRecipient:recipient.uniqueId];
```
charlesmchen, line 683: We now do some bookkeeping to track which group members have received a given message.

src/Messages/OWSMessageSender.m
```diff
@@ -678,14 +652,20 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
+    for (SignalRecipient *recipient in recipients) {
         // We don't need to send the message to ourselves...
-        if ([rec.uniqueId isEqualToString:[TSStorageManager localNumber]]) {
+        if ([recipient.uniqueId isEqualToString:[TSStorageManager localNumber]]) {
+            continue;
+        }
+        if ([message wasSentToRecipient:recipient.uniqueId]) {
+            // Skip recipients we have already sent this message to (on an
+            // earlier retry, perhaps).
+            DDLogInfo(@"Skipping group message recipient; already sent: %@", recipient.uniqueId);
```
charlesmchen, line 710: ...And we use that bookkeeping to reduce (but not completely avoid) double-sending.

src/Messages/OWSMessageSender.m
```diff
@@ -933,12 +913,10 @@ - (void)handleMismatchedDevices:(NSDictionary *)dictionary
 
 - (void)handleMessageSentLocally:(TSOutgoingMessage *)message
 {
-    [self saveMessage:message withState:TSOutgoingMessageStateSent];
```
charlesmchen, line 941: In https://github.com/WhisperSystems/SignalServiceKit/pull/167 we took a big step towards only marking messages as failed or succeeded in a single place and only when that state transition was "final".  It turns out there was more work to be done.  We now use the OWSSendMessageOperation's success and failure handlers to mark success and failure.

src/Messages/OWSMessageSender.m
```diff
@@ -1172,23 +1150,11 @@ - (TSWhisperMessageType)messageTypeForCipherMessage:(id<CipherMessage>)cipherMes
...
-    [message setSendingError:error];
-    [message save];
-}
-
 - (void)saveGroupMessage:(TSOutgoingMessage *)message inThread:(TSThread *)thread
 {
     if (message.groupMetaMessage == TSGroupMessageDeliver) {
-        [self saveMessage:message withState:message.messageState];
+        // TODO: Why is this necessary?
+        [message save];
```
charlesmchen, line 1237: I need to investigate more before I remove this change.

src/Messages/OWSMessageSender.m
```diff
@@ -1224,27 +1189,6 @@ - (void)handleStaleDevicesWithResponse:(NSData *)responseData
     });
 }
 
-- (void)processException:(NSException *)exception
-         outgoingMessage:(TSOutgoingMessage *)message
-                inThread:(TSThread *)thread
-{
-    DDLogWarn(@"%@ Got exception: %@", self.tag, exception);
-
-    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
```
charlesmchen, line 1238: I believe this method was redundant. It did two things:

* Save a new error message, which was never initialized, so this had no effect.
* Mark a message send as failed.  We want to do this in `OWSSendMessageOperation`'s failure handler, not here.

charlesmchen:
I guess it's worth stating that these changes are sensitive and need to be extensively tested.

src/Messages/OWSMessageSender.m
michaelkirk: I'll be really happy to get rid of this. =)

When this code was written, Android nor iOS had timestamp deduping in place, so it was a real problem.

Signal-Desktop is still missing timestamp deduping support, otherwise, I'd just say let's keep our code simple and blast it out to everyone.

So I think this approach is great for now. Maybe in the fully deduped future we can come back and simplify it.

charlesmchen:
@michaelkirk I've uncovered more work to be done here. Please hold off on reviewing this PR.

charlesmchen:
@michaelkirk I've uncovered more work to be done here. Please hold off on reviewing this PR.

charlesmchen:
@michaelkirk This PR is now ready for review.

src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -110,20 +187,163 @@ - (nullable NSString *)recipientIdentifier
...
+        case TSOutgoingMessageStateSent_OBSOLETE:
+        case TSOutgoingMessageStateDelivered_OBSOLETE:
+            OWSAssert(0);
+            return self.isExpiringMessage;
+    }
+}
+
+#pragma mark - Update Methods
+
+- (void)applyChangeToSelfAndLatestOutgoingMessage:(YapDatabaseReadWriteTransaction *)transaction
```
michaelkirk: The reason for this method may not be obvious outside of the context of this PR. Could you add a method comment to explain why it exists?

src/Messages/OWSMessageSender.m
```diff
@@ -697,26 +677,12 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
-
-            // At this point, all recipients must have failed.
-            // But we have all this verbose type checking because TOCFuture doesn't expose type information.
-            id lastError = errors.lastObject;
-            if ([lastError isKindOfClass:[TOCFuture class]]) {
-                TOCFuture *failedFuture = (TOCFuture *)lastError;
-                if (failedFuture.hasFailed) {
-                    id failureResult = failedFuture.forceGetFailure;
+            NSArray *groupSendFutures = (NSArray *)failure;
+            for (TOCFuture *groupSendFuture in groupSendFutures) {
```
michaelkirk: As I understand this change, we're now going to mark a group send as failed, if a single recipient fails to receive it. 

That has an obvious up side. The downside is that we'll confuse the situation where multiple recipients do get it but, e.g. the one unregistered group member, does not get it.

Is there a way to address this without going down the design-rabbit-hole of per-recipient delivery status? Just brainstorming... like wrapping the error and displaying it in the retry copy or footer status.

Even if not, I think what you have is a more conservative (better in this case) approach.

michaelkirk (approved):
LGTM

src/Messages/OWSMessageSender.m
charlesmchen: That's interesting. I'm inclined to leave the new state tracking which recipients have received the message.  I'm hoping it improves the experience when sending messages to very large groups (100 or 1,000 members) from bad connections, where the odds of sending failures is high - by tracking "sent recipients", we can "resume" the send from where the error took place.  WDYT?

michaelkirk: Sounds good.

src/Messages/Interactions/TSOutgoingMessage.m
charlesmchen: Done.

src/Messages/OWSMessageSender.m
charlesmchen: > The downside is that we'll confuse the situation where multiple recipients do get it but, e.g. the one unregistered group member, does not get it.

You raise an excellent point.  I thought about this overnight and came up with some further changes: https://github.com/WhisperSystems/SignalServiceKit/pull/173/commits/a7df1c18d0eb515bfe34ecb3fc1b05f313811573

* I've moved the "is retryable" flag from a separate boolean in the failure handler to being a property of the error.  This lets us propagate the flag from the inner loop to the outer loop.
* I've added a "error should be ignored for group sends" flag.  So far I'm only using it for the case you mentioned (sending group messages shouldn't appear to fail if one group member unregisters), but I wonder if we should use it for any other kind of error as well.
* I've also made a number of small bug fixes.  I'll call them out with inline comments.

src/Messages/OWSMessageSender.h
```diff
@@ -20,7 +20,23 @@ NS_ASSUME_NONNULL_BEGIN
  * but *sometimes* we don't want to retry when we know it's a terminal failure, so we allow the
  * caller to indicate this with isRetryable=NO.
  */
-typedef void (^RetryableFailureHandler)(NSError *_Nonnull error, BOOL isRetryable);
+typedef void (^RetryableFailureHandler)(NSError *_Nonnull error);
```
charlesmchen, line 23: This flag has moved to the `NSError` category below.

src/Messages/OWSMessageSender.m
```diff
@@ -536,16 +574,14 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
                 [self getRecipients:gThread.groupModel.groupMemberIds error:&error];
 
             if (recipients.count == 0) {
-                if (error) {
```
charlesmchen, line 567: Just some tidying up here.

src/Messages/OWSMessageSender.m
```diff
@@ -566,13 +602,17 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
                 ? self.storageManager.localNumber
                 : contactThread.contactIdentifier;
 
+            // If we block a user, don't send 1:1 messages to them. The UI
+            // should prevent this from occurring, but in some edge cases
+            // you might, for example, have a pending outgoing message when
+            // you block them.
```
charlesmchen, line 608: Added a comment.

src/Messages/OWSMessageSender.m
```diff
@@ -680,22 +721,29 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
     [completionFuture catchDo:^(id failure) {
-        // failure from toc_thenAll yeilds an array of failed Futures, rather than the future's failure.
+        // failure from toc_thenAll yields an array of failed Futures, rather than the future's failure.
         if ([failure isKindOfClass:[NSArray class]]) {
             NSArray *groupSendFutures = (NSArray *)failure;
             for (TOCFuture *groupSendFuture in groupSendFutures) {
                 if (groupSendFuture.hasFailed) {
                     id failureResult = groupSendFuture.forceGetFailure;
                     if ([failureResult isKindOfClass:[NSError class]]) {
-                        return failureHandler(failureResult, YES);
```
charlesmchen: Instead of marking all group failures as retry-able, I think we should propagate the original isRetryable flag...

src/Messages/OWSMessageSender.m
```diff
@@ -680,22 +721,29 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
+                        NSError *error = failureResult;
+                        // Some errors should be ignored when sending messages
+                        // to groups.  See discussion on
+                        // NSError (OWSMessageSender) category.
+                        if ([error shouldBeIgnoredForGroups]) {
```
charlesmchen, line 739: ...But there are some errors that we want to ignore when send messages to groups.

src/Messages/OWSMessageSender.m
```diff
@@ -831,13 +887,19 @@ - (void)sendMessage:(TSOutgoingMessage *)message
...
                 case 404: {
                     [self unregisteredRecipient:recipient message:message thread:thread];
                     NSError *error = OWSErrorMakeNoSuchSignalRecipientError();
                     // No need to retry if the recipient is not registered.
-                    return failureHandler(error, NO);
+                    [error setIsRetryable:NO];
+                    // If one member of a group deletes their account,
+                    // the group should ignore errors when trying to send
+                    // messages to this ex-member.
+                    [error setShouldBeIgnoredForGroups:YES];
```
charlesmchen, line 927: A key change.

src/Messages/OWSMessageSender.m
charlesmchen: @michaelkirk Hmm, looking over this last commit, I'm having second thoughts about whether moving the `isRetryable` flag from being a block parameter to a NSError property was a good idea.  Originally I thought it'd be a good idea for two reasons:

* I thought I might split it out into separate `isRetryableForContacts` and `isRetryableForGroups` flags, but my intuition was wrong - I don't think there aren't any scenarios where we want to retry for contacts but not for groups or vice versa.  
* It'd be consistent with the `shouldBeIgnoredForGroups` property.  But since we're using that in so few place,s who cares?

Anyway, I might revert most of this commit.

charlesmchen: Okay, I've thought it over and there is a solid reason to hang the `isRetryable` state on the error - we want to look at it when resolving possibly multiple errors for a group send into a single error.  If any of them are retryable, we want to retry.  I'm going to add another commit that does this.

src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -8,44 +8,93 @@
...
+                  expiresInSeconds:0];
 }
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
                          inThread:(nullable TSThread *)thread
                       messageBody:(nullable NSString *)body
                     attachmentIds:(NSMutableArray<NSString *> *)attachmentIds
 {
+    OWSAssert(0);
+
```
michaelkirk: Is this like a deprecation warning? Dare you just delete the method?

or at least add a comment.

src/Messages/OWSMessageSender.m
```diff
@@ -564,16 +574,14 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
...
-                    return;
-                } else {
+                if (!error) {
                     DDLogError(@"%@ Unknown error finding contacts", self.tag);
-                    // If not recipients were found, there's no reason to retry. It will just fail again.
-                    failureHandler(OWSErrorMakeFailedToSendOutgoingMessageError(), NO);
-                    return;
+                    error = OWSErrorMakeFailedToSendOutgoingMessageError();
                 }
+                // If not recipients were found, there's no reason to retry. It will just fail again.
```
michaelkirk: nit: copied typo, but :+1: on the tidying.

src/Messages/OWSMessageSender.m
```diff
@@ -594,13 +602,17 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
...
                 : contactThread.contactIdentifier;
 
+            // If we block a user, don't send 1:1 messages to them. The UI
+            // should prevent this from occurring, but in some edge cases
+            // you might, for example, have a pending outgoing message when
+            // you block them.
             OWSAssert(recipientContactId.length > 0);
             NSArray<NSString *> *blockedPhoneNumbers = _blockingManager.blockedPhoneNumbers;
             if ([blockedPhoneNumbers containsObject:recipientContactId]) {
-                DDLogInfo(@"%@ skipping 1:1 send to blocked contact: %@", self.tag, recipientContactId);
```
michaelkirk: nit: why remove this log? It's unlikely to be filling the logs up, and if it were I'd almost certainly want to know.

src/Messages/OWSMessageSender.m
```diff
@@ -700,36 +721,45 @@ - (void)groupSend:(NSArray<SignalRecipient *> *)recipients
...
-        return failureHandler(OWSErrorMakeFailedToSendOutgoingMessageError(), YES);
+        // If any of the group send errors are retryable, we want to retry.
+        // Therefore, prefer to propagate a retryable error.
+        if (firstRetryableError) {
+            return failureHandler(firstRetryableError);
+        } else if (firstNonRetryableError) {
+            return failureHandler(firstNonRetryableError);
+        } else {
+            // If we only received errors that we should ignore,
+            // consider this send a success.
```
michaelkirk: I don't think it's a big deal, but when you're in a group with *only* unregistered users, I expect we'll see the message successfully marked as "sent" even though we sent it to noone.

I don't think this is that big of a deal, since the sender will still see all the "unregistered user" warnings, they will likely not be that "surprised" that their message wasn't sent, but it's a little weird.

Would it be straight forward to do something like:
```
 if (message.sentRecipients.count == 0) {
    // do we need to mark it as non-retryable?
    NSError *error = OWSErrorMakeNoValidRecipientsError()
    failureHandler(errror)
}
```

edit: To restate, if this isn't as straight forward to address as I think, it might not be worth addressing.

src/Messages/OWSMessageSender.m
```diff
@@ -952,7 +995,7 @@ - (void)handleMessageSentLocally:(TSOutgoingMessage *)message
 
 - (void)handleMessageSentRemotely:(TSOutgoingMessage *)message sentAt:(uint64_t)sentAt
 {
-    [self saveMessage:message withState:TSOutgoingMessageStateDelivered];
+    [message updateWithWasDelivered];
```
michaelkirk, line 1008: Not your bug, but shouldn't this actually just be "sent"?

We have no way of knowing if it was delivered to Bob yet, right?

michaelkirk (approved):
Just minor feedback. LGTM

charlesmchen:
> I don't think it's a big deal, but when you're in a group with only unregistered users, I expect we'll see the message successfully marked as "sent" even though we sent it to no-one.

It's a strange edge case; I'm not sure what the desired behavior should be, so I'll defer to you.

> Would it be straight forward to do something like:

Done.

charlesmchen:
> nit: copied typo, but 👍 on the tidying.

Done.

charlesmchen:
> nit: why remove this log? It's unlikely to be filling the logs up, and if it were I'd almost certainly want to know.

Done.

charlesmchen:
> Is this like a deprecation warning? Dare you just delete the method?

I was going to try to deprecate some of these initializers, but it's just not worth it. Reverted.

charlesmchen:
> Not your bug, but shouldn't this actually just be "sent"?

I'm going to add a card so that we discuss this.

PR #174, opened 2017-04-12
Update SignalAttachment to allow arbitrary attachments.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1976

michaelkirk (approved):
LGTM

PR #175, opened 2017-04-13
Reduce usage of contacts intersection endpoint.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1977

PTAL @michaelkirk

### DIFF
--- a/src/Contacts/ContactsUpdater.m
+++ b/src/Contacts/ContactsUpdater.m
@@ -14,6 +14,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSString *const kContactsUpdaterErrorDomain = @"kContactsUpdaterErrorDomain";
+const NSInteger kContactsUpdaterRateLimitErrorCode = 413;
+
 @implementation ContactsUpdater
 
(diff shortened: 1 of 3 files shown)


src/Contacts/ContactsUpdater.m
```diff
@@ -14,6 +14,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSString *const kContactsUpdaterErrorDomain = @"kContactsUpdaterErrorDomain";
```
michaelkirk: I understand the argument for using more fine grained error domains, and if you want to go that direction that's fine, but it would be more consistent with the rest of the code base to add an error to the OWSError class

Going forward with what you have here will mean we need to be more careful about verifying the ErrorDomain whenever we're inspecting error codes (which we *should*, but probably *are not* always doing) lest we collide on common error codes like "413".

michaelkirk (approved):
LGTM

src/Contacts/ContactsUpdater.m
charlesmchen: Done.

PR #177, opened 2017-04-13
Charlesmchen/honor attachment filenames

charlesmchen:
* Send filenames with outgoing attachments, if possible.
* Store incoming attachment filenames, if present.
* Write attachment data to disk using the attachment's filename, if possible.

PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1978

protobuf/OWSSignalServiceProtos.proto
```diff
@@ -39,12 +39,12 @@ message Content {
 message CallMessage {
   message Offer {
     optional uint64 id          = 1;
-    optional string sessionDescription = 2;
```
michaelkirk: I think I made this change myself - sorry for not documenting it, but clobbering the NSObject `description` method seemed wrong.

src/Util/MIMETypeUtil.m
```diff
@@ -277,7 +277,65 @@ + (BOOL)isAudio:(NSString *)contentType {
 
 + (NSString *)filePathForAttachment:(NSString *)uniqueId
                          ofMIMEType:(NSString *)contentType
-                           inFolder:(NSString *)folder {
+                           filename:(nullable NSString *)filename
+                           inFolder:(NSString *)folder
+{
+    NSString *kDefaultFileExtension = @"bin";
```
charlesmchen, line 267: We repeatedly call this method to determine what the filepath for a given attachment is. Instead, we should call it once, and store the file path on the attachment model.  That would make us robust to changes to this method's behavior.

src/Util/MIMETypeUtil.m
```diff
@@ -277,7 +277,65 @@ + (BOOL)isAudio:(NSString *)contentType {
...
+
+    if (filename.length > 0) {
+        NSString *normalizedFilename =
+            [filename stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+        // Ensure that the filename is a valid filesystem name,
+        // replacing invalid characters with an underscore.
+        for (NSCharacterSet *invalidCharacterSet in @[
+                 [NSCharacterSet whitespaceCharacterSet],
```
michaelkirk: dare we `whiteSpacesAndNewLinesCharacterSet`

src/Util/MIMETypeUtil.m
```diff
@@ -277,7 +277,65 @@ + (BOOL)isAudio:(NSString *)contentType {
...
+                 [NSCharacterSet newlineCharacterSet],
+                 [NSCharacterSet illegalCharacterSet],
+                 [NSCharacterSet controlCharacterSet],
+                 [NSCharacterSet characterSetWithCharactersInString:@"<>|\\:()&;?*"],
```
michaelkirk: Other things that come to mind "/", "~", "."

src/Messages/Interactions/TSMessage.h
```diff
@@ -23,6 +23,7 @@ typedef NS_ENUM(NSInteger, TSGroupMetaMessage) {
 @interface TSMessage : TSInteraction
 
 @property (nonatomic, readonly) NSMutableArray<NSString *> *attachmentIds;
+// A map of attachment id-to-filename.
```
michaelkirk: I think this comment was intended for TSOutgoingMessage.h

src/Util/MIMETypeUtil.m
michaelkirk: agree

michaelkirk (approved):
LGTM

protobuf/OWSSignalServiceProtos.proto
charlesmchen: Reverted.

src/Util/MIMETypeUtil.m
charlesmchen: Done.

src/Util/MIMETypeUtil.m
charlesmchen: I've added forward-slash / and tilde ~, but I'm omitting period since it's so widely used in filenames like: "Budget.latest.xls" or whatever.  I'll do explicit checks for filenames like "." and "..".

charlesmchen: Hmm, maybe I'll strip leading periods too.

michaelkirk: perfect, thanks!

src/Messages/Interactions/TSMessage.h
charlesmchen: Done.

src/Util/MIMETypeUtil.m
charlesmchen: I've added a card.

PR #179, opened 2017-04-14
De-bounce the prekey checks.

charlesmchen:
After we talked yesterday, I realized that there's a safe, easy to way to de-bounce prekey checks without introducing any new state.  

* We optimistically mark the prekeys as "checked".
* We mark them as "unchecked" if the check or prekey registration fails in any way.  

PTAL @michaelkirk

### DIFF
--- a/src/Account/TSPreKeyManager.h
+++ b/src/Account/TSPreKeyManager.h
@@ -26,7 +26,7 @@ typedef NS_ENUM(NSInteger, RefreshPreKeysMode) {
                         success:(void (^)())successHandler
                         failure:(void (^)(NSError *error))failureHandler;
 
-+ (void)refreshPreKeys;
++ (void)checkPreKeys;
 
 + (void)checkPreKeysIfNecessary;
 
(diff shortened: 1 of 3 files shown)


src/Account/TSPreKeyManager.h
```diff
@@ -26,7 +26,7 @@ typedef NS_ENUM(NSInteger, RefreshPreKeysMode) {
                         success:(void (^)())successHandler
                         failure:(void (^)(NSError *error))failureHandler;
 
-+ (void)refreshPreKeys;
++ (void)checkPreKeys;
```
charlesmchen, line 29: The name "refresh" was misleading; it suggested that this method would always register new prekeys.  This method checks whether we need to register and only registers if necessary.

src/Account/TSPreKeyManager.m
```diff
@@ -241,11 +269,6 @@ + (void)refreshPreKeys {
                     DDLogDebug(@"%@ Not updating prekeys.", self.tag);
                 }
             }
-            
-            // Update the prekey check timestamp on success.
-            dispatch_async(TSPreKeyManager.prekeyQueue, ^{
-                lastPreKeyCheckTimestamp = [NSDate date];
-            });
```
charlesmchen, line 249: Previously we waited until the check (but not prekey) registration succeeded to mark the check as complete.  

* We were marking the check complete after stage 1 of the check but not after completion of the optional stage 2 of the check, so that wasn't great.
* Due to asynchronicity, this wouldn't de-bounce check requests that followed fast on each others' heels.
* We also want to re-check if any part of the registration fails.

michaelkirk (approved):
LGTM

PR #180, opened 2017-04-14
Avoid lost messages by acknowledges message receipt after the message…

charlesmchen:
… is processed.

Another change that will require careful testing.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSMessagesManager.h
+++ b/src/Messages/TSMessagesManager.h
@@ -18,6 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol ContactsManagerProtocol;
 @protocol OWSCallMessageHandler;
 
+typedef void (^MessageManagerCompletionBlock)();
+
 @interface TSMessagesManager : NSObject
 
 - (instancetype)init NS_UNAVAILABLE;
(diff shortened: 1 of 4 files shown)


src/Messages/TSMessagesManager.h
```diff
@@ -27,18 +29,8 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)envelope;
-
-/**
- * Processes all kinds of incoming envelopes with a data message, along with any attachments.
- *
- * @returns
- *   If an incoming message is created, it will be returned. If it is, for example, a group update,
- *   no incoming message is created, so nil will be returned.
- */
-- (TSIncomingMessage *)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)envelope
```
charlesmchen, line 39: This method was only used internally, so I've removed it from the header.

src/Messages/TSMessagesManager.m
```diff
@@ -241,14 +259,17 @@ - (void)handleSecureMessageAsync:(OWSSignalServiceProtosEnvelope *)messageEnvelo
                         [TSErrorMessage missingSessionWithEnvelope:messageEnvelope withTransaction:transaction];
                     [errorMessage saveWithTransaction:transaction];
                 }];
+                DDLogError(@"Skipping message envelope for unknown session.");
+                completion(nil);
```
charlesmchen, line 263: There were lots of places in this method were we never called the completion handler.

src/Messages/TSMessagesManager.m
```diff
@@ -297,6 +318,7 @@ - (void)handlePreKeyBundleAsync:(OWSSignalServiceProtosEnvelope *)preKeyEnvelope
         NSData *encryptedData = preKeyEnvelope.hasContent ? preKeyEnvelope.content : preKeyEnvelope.legacyMessage;
         if (!encryptedData) {
             DDLogError(@"Skipping message envelope which had no encrypted data");
+            completion(nil);
```
charlesmchen, line 321: Same for this method.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/1985

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -381,15 +379,22 @@ - (void)processWebSocketRequestMessage:(WebSocketRequestMessage *)message {
 
         if (!decryptedPayload) {
             DDLogWarn(@"Failed to decrypt incoming payload or bad HMAC");
```
michaelkirk, line 381: I think we should display a user facing error here, it's arguably out of scope, so feel free to add a new card.

michaelkirk (approved):
LGTM

src/Network/WebSockets/TSSocketManager.m
charlesmchen: Added a card.

PR #183, opened 2017-04-19
Fix outgoing message status for messages sent from linked devices.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/SignalServiceKit/pull/173#discussion_r111617481

IMO these messages need to show up immediately as _delivered_ (and not just _sent_), or they'll never transition to _delivered_.  The bug was that I was _only_ marking them as delivered and not as sent too.


michaelkirk (approved):
LGTM

PR #184, opened 2017-04-19
Do not try to resend unsaved outgoing messages when accepting a safet…

charlesmchen:
…y number change.

This is not a change in behavior.  It fixes a regression introduced by the "outgoing messages state" changes.  

We don't persist most outgoing group messages (TSGroupMessageNew, TSGroupMessageUpdate, TSGroupMessageQuit), so we can't retry them after accepting a safety number change.  Which is a shame, because it'd be nice to resend them.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -391,6 +391,7 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
             success:(void (^)())successHandler
             failure:(void (^)(NSError *error))failureHandler
 {
+    OWSAssert(message);
     AssertIsOnMainThread();
 
     [message updateWithMessageState:TSOutgoingMessageStateAttemptingOut];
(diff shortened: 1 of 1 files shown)


src/Messages/OWSMessageSender.m
```diff
@@ -523,13 +524,27 @@ - (void)resendMessageFromKeyError:(TSInvalidIdentityKeySendingErrorMessage *)err
...
     //  1.) succeed and create a new successful message in the thread or...
     //  2.) fail and create a new identical error message in the thread.
     [errorMessage remove];
 
+    // The failedMessageId might be nil for transient, unsaved outgoing messages.
+    // See [TSOutgoingMessage saveWithTransaction:] for details of which messages
+    // we do not save.
+
+    if (!failedMessageId) {
+        return;
```
charlesmchen, line 542: This is the key change.

michaelkirk (approved):
LGTM and created a card to address the deeper issue.

PR #185, opened 2017-04-19
Do not retry fatal message send errors.

charlesmchen:
PTAL @michaelkirk


michaelkirk:
rebased and merged in 0bbe73e

PR #196, opened 2017-05-03
Properly handle "work", "home", and "other" labels

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/src/Contacts/Contact.m
+++ b/src/Contacts/Contact.m
@@ -72,10 +72,14 @@ NS_ASSUME_NONNULL_BEGIN
         if ([phoneNumberField.value isKindOfClass:[CNPhoneNumber class]]) {
             CNPhoneNumber *phoneNumber = (CNPhoneNumber *)phoneNumberField.value;
             [phoneNumbers addObject:phoneNumber.stringValue];
-            if ([phoneNumberField.label isEqualToString:CNLabelPhoneNumberMobile]) {
-                phoneNumberTypeMap[phoneNumber.stringValue] = @(OWSPhoneNumberTypeMobile);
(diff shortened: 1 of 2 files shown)


src/Contacts/Contact.m
```diff
@@ -86,7 +90,13 @@ - (instancetype)initWithContact:(CNContact *)contact
                 phoneNumberTypeMap[phoneNumber.stringValue] = @(OWSPhoneNumberTypeOtherFAX);
             } else if ([phoneNumberField.label isEqualToString:CNLabelPhoneNumberPager]) {
                 phoneNumberTypeMap[phoneNumber.stringValue] = @(OWSPhoneNumberTypePager);
+            } else if ([phoneNumberField.label isEqualToString:CNLabelOther]) {
+                phoneNumberTypeMap[phoneNumber.stringValue] = @(OWSPhoneNumberTypeOther);
             } else {
+                // TODO this should only be hit if the user has a custom label, which ideally we'd expose rather than
+                // unknown. I think the thing to do is to get rid of OWSPhoneNumberType*, and simply localize the label
+                // in this method - uising the custom label when available.
+                // Maybe that means adding a "label" to PhoneNumber class and parsing the numbers in this class.
```
charlesmchen, line 99: Interesting. We can definitely get rid of the enum. We should localize the system-defined labels though, right?

charlesmchen (approved):
LGTM

src/Contacts/Contact.m
michaelkirk: Yeah, I couldn't find any sort of built in localization for these.

PR #199, opened 2017-05-03
Temporary change to improve read receipt logging.

charlesmchen:
PTAL @michaelkirk 

We'll revert this before the next release, but it should allow us to fix this issue.


michaelkirk (approved):
LGTM

PR #200, opened 2017-05-04
Rework handling of phone number names.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/Contact.m
+++ b/src/Contacts/Contact.m
@@ -14,7 +14,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface Contact ()
 
-@property (readonly, nonatomic) NSMutableDictionary<NSString *, NSNumber *> *phoneNumberTypeMap;
+@property (readonly, nonatomic) NSMutableDictionary<NSString *, NSString *> *phoneNumberNameMap;
 
 @end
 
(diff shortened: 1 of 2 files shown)


src/Contacts/Contact.m
```diff
@@ -67,45 +65,58 @@ - (instancetype)initWithSystemContact:(CNContact *)contact
...
-                phoneNumberTypeMap[phoneNumber.stringValue] = @(OWSPhoneNumberTypeUnknown);
+                phoneNumberNameMap[phoneNumber.stringValue]
+                    = NSLocalizedString(@"PHONE_NUMBER_TYPE_OTHER", @"Label for 'Other' phone numbers.");
+            } else if (phoneNumberField.label.length > 0 && ![phoneNumberField.label hasPrefix:@"_$"]) {
+                // We'll reach this case for:
+                //
+                // * User-defined custom labels, which we want to display.
+                // * Labels like "_$!<CompanyMain>!$_", which I'm guessing are synced from other platforms.
+                //   We don't want to display these labels. Even some of iOS' default labels (like Radio) show
+                //   up this way.
```
charlesmchen, line 110: We could parse these with a regular expression, but I feel like this is an edge case we're already handling pretty well. I'm wary of relying on the details of how Apple handles these other labels.

michaelkirk (approved):
LGTM

PR #202, opened 2017-05-05
Charlesmchen/contacts sync

charlesmchen:
PTAL @michaelkirk


src/Contacts/SignalAccount.h
```diff
@@ -0,0 +1,51 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
```
michaelkirk, line 5: Was this just a straight-up move from Signal-iOS?

michaelkirk (approved):
nice, LGTM

src/Contacts/SignalAccount.h
charlesmchen: Yup!

PR #204, opened 2017-05-05
Auto-rejoin groups by emitting and responding to “request group info”…

charlesmchen:
… messages.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/DeviceSyncing/OWSSyncGroupsRequestMessage.m
+++ b/src/Messages/DeviceSyncing/OWSSyncGroupsRequestMessage.m
@@ -1,0 +1,73 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSSyncGroupsRequestMessage.h"
+#import "NSDate+millisecondTimeStamp.h"
+#import "OWSSignalServiceProtos.pb.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSSyncGroupsRequestMessage ()
+
+@property (nonatomic) NSData *groupId;
+
+@end
+
+#pragma mark -
+
+@implementation OWSSyncGroupsRequestMessage
+
+- (instancetype)initWithThread:(nullable TSThread *)thread groupId:(NSData *)groupId
+{
+    self = [super initWithTimestamp:[NSDate ows_millisecondTimeStamp] inThread:thread];
+    if (!self) {
+        return self;
+    }
+
+    OWSAssert(groupId.length > 0);
+    _groupId = groupId;
+
+    return self;
+}
+
+- (void)saveWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    // override superclass with no-op.
+    //
+    // There's no need to save this message, since it's not displayed to the user.
+}
+
+- (BOOL)shouldSyncTranscript
+{
+    return NO;
+}
+
+- (BOOL)isLegacyMessage
+{
+    return NO;
+}
+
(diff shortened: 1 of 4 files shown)


src/Messages/TSMessagesManager.m
```diff
@@ -490,7 +492,28 @@ - (void)handleIncomingEnvelope:(OWSSignalServiceProtosEnvelope *)incomingEnvelop
...
+            OWSSyncGroupsRequestMessage *syncGroupsRequestMessage =
+                [[OWSSyncGroupsRequestMessage alloc] initWithThread:thread groupId:groupId];
+            [self.messageSender sendMessage:syncGroupsRequestMessage
+                success:^{
+                    DDLogInfo(@"%@ Successfully sent Request Group Info message.", self.tag);
+                }
+                failure:^(NSError *error) {
+                    DDLogError(@"%@ Failed to send Request Group Info message with error: %@", self.tag, error);
+                }];
+
```
charlesmchen, line 516: @michaelkirk We could go a step further and persist the message and re-process it if/when we get the solicited group update. It'd be a nice refinement but doesn't seem essential, and would introduce a certain amount of surface area for bugs.

charlesmchen:
This introduces the risk that users will be auto-joined to groups they have deleted (these get deleted from the data store, right?).  If any of the group members didn't get your "quit group" message, they'll send you group messages, you'll send them a "request group info" message, they'll send it, you'll be back in the group.  I think this is acceptable. "quit group" messages should be usually delivered.

charlesmchen:
PTAL @michaelkirk

src/Messages/DeviceSyncing/OWSSyncGroupsRequestMessage.m
```diff
@@ -0,0 +1,73 @@
...
+}
+
+- (BOOL)shouldSyncTranscript
+{
+    return NO;
+}
+
+- (BOOL)isLegacyMessage
+{
+    return NO;
```
michaelkirk: I think this would be the first non-legacy message that is not a sync message.

In theory all the clients should be able to handle this by now, but I think it warrants some extensive testing across other clients.

Alternatively, I think you could make it a legacy message and use the existing `super` implementation of `buildPlainTextData`.

src/Messages/TSMessagesManager.m
```diff
@@ -737,6 +788,30 @@ - (TSIncomingMessage *)handleReceivedEnvelope:(OWSSignalServiceProtosEnvelope *)
...
     [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
       if (groupId) {
+          if (dataMessage.group.type == OWSSignalServiceProtosGroupContextTypeRequestInfo) {
+              DDLogInfo(@"Received 'Request Group Info' message for group: %@", groupId);
+
+              TSGroupModel *emptyModelToFillOutId =
+                  [[TSGroupModel alloc] initWithTitle:nil memberIds:nil image:nil groupId:dataMessage.group.id];
+              TSGroupThread *gThread =
+                  [TSGroupThread threadWithGroupModel:emptyModelToFillOutId transaction:transaction];
+              if (gThread) {
```
michaelkirk: nit: else error log//assert?

michaelkirk (approved):
LGTM

src/Messages/TSMessagesManager.m
charlesmchen: Done.

src/Messages/DeviceSyncing/OWSSyncGroupsRequestMessage.m
charlesmchen: Okay, we'll do it the legacy way.

PR #205, opened 2017-05-08
Always keep at least 3 old signed prekeys (accepted or otherwise).

michaelkirk:
If the keys weren't marked as accepted, we risk deleting keys which may
still be relevant.

PTAL @charlesmchen

### DIFF
--- a/src/Account/TSPreKeyManager.m
+++ b/src/Account/TSPreKeyManager.m
@@ -355,6 +355,8 @@ static const CGFloat kSignedPreKeyUpdateFailureMaxFailureDuration = 10 * 24 * 60
             return [left.generatedAt compare:right.generatedAt];
         }];
 
+        NSUInteger oldSignedPreKeyCount = oldSignedPrekeys.count;
+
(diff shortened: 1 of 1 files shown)


src/Account/TSPreKeyManager.m
```diff
@@ -383,6 +390,7 @@ + (void)clearSignedPreKeyRecordsWithKeyId:(NSNumber *)keyId success:(void (^_Nul
                 [dateFormatter stringFromDate:signedPrekey.generatedAt],
                 signedPrekey.wasAcceptedByService);
 
+            oldSignedPreKeyCount--;
```
charlesmchen, line 393: We also discussed using a different (longer) `kSignedPreKeysDeletionTime` for signed prekeys that were generated before February. Should we do that too?

charlesmchen (approved):
LGTM

michaelkirk:
In https://github.com/WhisperSystems/SignalServiceKit/commit/311206918fd2ce8948d35041d74a26c24733742a

PR #206, opened 2017-05-08
Try to deduce attachment MIME type from the file extension if possible.

charlesmchen:
PTAL @michaelkirk


src/Util/MIMETypeUtil.m
```diff
@@ -1546,6 +1561,1006 @@ + (NSDictionary *)genericMIMETypesToExtensionTypes
...
+    return [self genericExtensionTypesToMIMETypes][fileExtension];
+}
+
++ (NSDictionary *)genericExtensionTypesToMIMETypes
+{
+    static NSDictionary *result = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        result = @{
+            @"123" : @"application/vnd.lotus-1-2-3",
```
charlesmchen, line 1577: I might move this mapping and the mapping above to .txt resources.

src/Util/MIMETypeUtil.m
```diff
@@ -1546,18 +1561,1023 @@ + (NSDictionary *)genericMIMETypesToExtensionTypes
...
+}
+
 + (nullable NSString *)fileExtensionForMIMETypeViaLookup:(NSString *)mimeType
 {
     return [[self genericMIMETypesToExtensionTypes] objectForKey:mimeType];
 }
 
 + (nullable NSString *)fileExtensionForMIMEType:(NSString *)mimeType
 {
-    // Try to deduce the file extension by converting to a UTI type.
```
charlesmchen, line 1556: This is an important change. Now that I've realized that the UTI type-based lookup for file extensions isn't 100% reliable, I'm going to prefer to determine the attachment file extension using lookup. This places a greater responsibility on our lookup mapping to be reliable.

michaelkirk (commented):
This is some kind of unfortunate code in that it's hard to review a list like that.

Also that list of mappings is *almost but not quite* an inverse of another giant list.

To step back a minute, what is the root reason for this commit, in terms of user experience?

charlesmchen:
> Also that list of mappings is almost but not quite an inverse of another giant list.

Yup, we can just include the list once because it's a many-to-one mapping in which we deliberately want to favor certain members of the many in the inverse mapping. 

'image/jpeg' MIME type maps to .jpg and .jpeg., but we want to make sure that we use .jpg (not .jpeg) as the favored file extension.

> To step back a minute, what is the root reason for this commit, in terms of user experience?

The primary motivation is this.  Signal's attachment all require a MIME type.  We always have a UTI type for all outgoing attachments, and will sometimes (except in the paste case) have a filename.  If we have a filename, it is far more reliable to determine the MIME type from the file extension than from the UTI type, not least because there are far fewer UTI types so the vast majority of MME types & file extensions have no corresponding UTI type.  

> This is some kind of unfortunate code in that it's hard to review a list like that.

I'd argue that the mapping itself doesn't need to be reviewed; it's directly derived the Apache mapping which should be very reliable.  

See: https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

It's the (existing) inverse mapping (where we need to express our preference for .jpg over .jpeg, for example) that is more subjective and where I resolved conflicts myself by hand.

michaelkirk:
compelled!

LGTM

PR #208, opened 2017-05-09
Rework the “disappearing messages” logic.

charlesmchen:
* Convert `OWSDisappearingMessagesJob` to a singleton since its work is expensive.
* Ensure `OWSDisappearingMessagesJob` does all of its work on a private queue, off the main thread.
* Rework scheduling of `OWSDisappearingMessagesJob` runs to use a NSTimer, so that we can cancel redundant scheduled runs.
* Move NSTimer+OWS category to SSK.
* Fix bug wherein some disappearing messages never got deleted.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSDisappearingMessagesJob.m
+++ b/src/Messages/OWSDisappearingMessagesJob.m
@@ -52,7 +52,21 @@ NS_ASSUME_NONNULL_BEGIN
 
     OWSSingletonAssert();
 
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(applicationDidBecomeActive:)
+                                                 name:UIApplicationDidBecomeActiveNotification
+                                               object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(applicationWillResignActive:)
+                                                 name:UIApplicationWillResignActiveNotification
+                                               object:nil];
+
     return self;
+}
+
+- (void)dealloc
+{
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
 + (dispatch_queue_t)serialQueue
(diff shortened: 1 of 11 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2087

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -69,23 +99,19 @@ - (void)runLoop
...
             @"%@ No more expiring messages. Setting next check %u seconds into the future", self.tag, delaySeconds);
-        [self runBy:now + delaySeconds * 1000];
+        [self runByDate:[NSDate ows_dateWithMillisecondsSince1970:now + delaySeconds * 1000]];
         return;
     }
 
     uint64_t nextExpirationAt = [nextExpirationTimestampNumber unsignedLongLongValue];
-    uint64_t runByMilliseconds;
-    if (nextExpirationAt < now + 1000) {
-        DDLogWarn(@"%@ Next run requested at %llu, which is too soon. Delaying by 1 sec to prevent churn",
```
charlesmchen, line 79: `runByDate` already ensures that we don't run more often than once per second, so this logic is no longer necessary.

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -69,23 +99,19 @@ - (void)runLoop
...
-        runByMilliseconds = nextExpirationAt;
-    }
+    [self runByDate:[NSDate ows_dateWithMillisecondsSince1970:MAX(nextExpirationAt, now)]];
+}
 
-    DDLogVerbose(@"%@ Requesting next expiration to run by: %llu", self.tag, nextExpirationAt);
-    [self runBy:runByMilliseconds];
++ (void)setExpirationForMessage:(TSMessage *)message
+{
+    dispatch_async(self.serialQueue, ^{
```
charlesmchen, line 124: These public class method make it easier to use the singleton and enforce the invariant that all work done by this class is done on its private queue.

src/Util/NSTimer+OWS.h
```diff
@@ -0,0 +1,19 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
```
charlesmchen, line 2: This is a straight move.

charlesmchen:
I pushed another commit:


* Disable “disappearing messages” job when inactive.

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -69,23 +113,19 @@ - (void)runLoop
         unsigned int delaySeconds = (10 * 60); // 10 minutes.
         DDLogDebug(
             @"%@ No more expiring messages. Setting next check %u seconds into the future", self.tag, delaySeconds);
-        [self runBy:now + delaySeconds * 1000];
+        [self runByDate:[NSDate ows_dateWithMillisecondsSince1970:now + delaySeconds * 1000]];
```
michaelkirk: Is this `if` block still necessary? It seems redundant with intent of kMaxDelaySeconds.

michaelkirk (approved):
LGTM

src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: I've added a `runLater` method here to clarify the intention.

PR #211, opened 2017-05-11
Charlesmchen/manual censorship circumvention

charlesmchen:
Not for this release (v2.12).

PTAL @michaelkirk

### DIFF
--- a/src/Network/API/TSNetworkManager.m
+++ b/src/Network/API/TSNetworkManager.m
@@ -27,7 +27,7 @@ typedef void (^failureBlock)(NSURLSessionDataTask *task, NSError *error);
     static TSNetworkManager *sharedMyManager = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        OWSSignalService *signalService = [[OWSSignalService alloc] init];
+        OWSSignalService *signalService = [OWSSignalService sharedInstance];
         sharedMyManager = [[self alloc] initWithSignalService:signalService];
     });
     return sharedMyManager;
--- a/src/Network/API/TSNetworkManager.m
+++ b/src/Network/API/TSNetworkManager.m
@@ -14,7 +14,6 @@ NSString *const TSNetworkManagerDomain = @"org.whispersystems.signal.networkMana
 
 @interface TSNetworkManager ()
 
-@property (nonatomic, readonly, strong) OWSSignalService *signalService;
 typedef void (^failureBlock)(NSURLSessionDataTask *task, NSError *error);
 
 @end
@@ -27,21 +26,18 @@ typedef void (^failureBlock)(NSURLSessionDataTask *task, NSError *error);
     static TSNetworkManager *sharedMyManager = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        OWSSignalService *signalService = [OWSSignalService sharedInstance];
-        sharedMyManager = [[self alloc] initWithSignalService:signalService];
+        sharedMyManager = [[self alloc] initDefault];
     });
     return sharedMyManager;
 }
 
 
-- (instancetype)initWithSignalService:(OWSSignalService *)signalService
+- (instancetype)initDefault
 {
     self = [super init];
     if (!self) {
         return self;
     }
-
-    _signalService = signalService;
 
     OWSSingletonAssert();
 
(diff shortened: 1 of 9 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2096

src/Network/OWSSignalService.m
```diff
@@ -28,24 +52,105 @@ - (instancetype)init
...
+    }
+
+    [self updateIsCensorshipCircumventionActive];
+}
+
+- (BOOL)isCensorshipCircumventionManuallyActivated
+{
+    OWSAssert([NSThread isMainThread]);
+
+    return [[NSUserDefaults.standardUserDefaults
```
michaelkirk: Other than logging, which has to be set up before the DB is readable, I think this is the only preference stored outside of the DB, which is a little unfortunate.

But all the preferences stuff currently lives outside of SSK, so I can understand why it wasn't quite straightforward to store this there.

So, should we expose user preferences via a protocol? Move them to SSK? Not worry about it?

src/Network/OWSSignalService.m
```diff
@@ -28,24 +52,105 @@ - (instancetype)init
...
+    }
+
+    [[NSNotificationCenter defaultCenter]
+        postNotificationName:kNSNotificationName_IsCensorshipCircumventionActiveDidChange
+                      object:nil
+                    userInfo:nil];
+}
+
+- (BOOL)isCensorshipCircumventionActive
+{
```
michaelkirk, line 150: `OWSAssert([NSThread isMainThread])`?

src/Network/WebSockets/TSSocketManager.m
```diff
@@ -286,17 +278,28 @@ - (void)setStatus:(SocketStatus)status {
...
+            _state = state;
             [socket open];
             return;
         }
     }
 
-    _status = status;
+    _state = state;
+
+    [self notifyStatusChange];
```
michaelkirk, line 291: :+1: to this over KVO

michaelkirk (approved):
LGTM

src/Network/OWSSignalService.m
charlesmchen: Access to the property is `@synchronized` because this might be accessed on any thread.

src/Network/OWSSignalService.m
charlesmchen: It didn't seem like a secret so I wasn't going to go to all of that trouble. WDYT?

src/Network/WebSockets/TSSocketManager.m
charlesmchen: I abhor KVO.

charlesmchen:
@michaelkirk I pushed another commit. PTAL

src/Network/OWSSignalService.m
michaelkirk: Sure, just wanted to make sure you weren't breaking ground here accidentally.

src/Network/API/TSNetworkManager.m
```diff
@@ -59,7 +55,7 @@ - (void)makeRequest:(TSRequest *)request
     void (^failure)(NSURLSessionDataTask *task, NSError *error) =
         [TSNetworkManager errorPrettifyingForFailureBlock:failureBlock];
 
-    AFHTTPSessionManager *sessionManager = self.signalService.HTTPSessionManager;
```
michaelkirk, line 62: Why get rid of this dependency injection?

src/Network/OWSCensorshipConfiguration.m
```diff
@@ -21,10 +21,10 @@ - (NSString *)frontingHost:(NSString *)e164PhonNumber
             domain = self.censoredCountryCodes[countryCode];
         }
     }
-    
-    // Fronting should only be used for countries specified in censoredCountryCodes,
-    // all of which have a domain specified.
-    OWSAssert(domain);
```
michaelkirk, line 27: were we hitting this assert?

charlesmchen: Yup, check out the changes to the comment here.  I'll tell you more at standup.

src/Network/API/TSNetworkManager.m
charlesmchen: Because there is a dependency cycle.

`AppDelegate` creates `TSNetworkManager` which requires `OWSSignalService` which requires `TSAccountManager` which requires `TSNetworkManager`.  

What I'd like to do is lazy-initialize `OWSSignalService` but there isn't a good way to do that without duplicating/complicating much of its logic or running afoul of synchronization issues.

Happy to discuss.

src/Network/OWSSignalService.m
charlesmchen: I added some more properties and moved to storing them in `TSStorageManager`.

charlesmchen:
I pushed another commit. PTAL @michaelkirk

PR #213, opened 2017-05-11
hashable method to detect when contact has changed

michaelkirk:
Used by https://github.com/WhisperSystems/Signal-iOS/pull/2101

PTAL @charlesmchen

### DIFF
--- a/src/Contacts/Contact.m
+++ b/src/Contacts/Contact.m
@@ -3,6 +3,7 @@
 //
 
 #import "Contact.h"
+#import "Cryptography.h"
 #import "PhoneNumber.h"
 #import "SignalRecipient.h"
 #import "TSAccountManager.h"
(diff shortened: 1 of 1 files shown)


src/Contacts/Contact.m
```diff
@@ -281,6 +281,22 @@ - (NSString *)nameForPhoneNumber:(NSString *)recipientId
     return value;
 }
 
+- (NSUInteger)hash
+{
+    NSUInteger hash = 1825038313 ^ self.fullName.hash;
+
+    // NSData.hash appears not to change even when the underlying bytes change.
+    // maybe it's built on address?
+    hash = hash ^ self.cnContact.thumbnailImageData.description.hash;
```
charlesmchen: Maybe use `Cryptography` function to compute hash and then hash those bytes?

charlesmchen: `Contact` has an `emails` property... do we use that anywhere?  I think we do - shouldn't we include that in this hash so that changes to emails update the current contacts?

charlesmchen (approved):
Otherwise LGTM

src/Contacts/Contact.m
michaelkirk: > Contact has an emails property... do we use that anywhere? I think we do - shouldn't we include that in this hash so that changes to emails update the current contacts?

Good call.

michaelkirk: > Maybe use Cryptography function to compute hash and then hash those bytes?
  
Done!

PR #214, opened 2017-05-12
Only reply to group info requester.

charlesmchen:
PTAL @michaelkirk


michaelkirk (commented):
LGTM

charlesmchen:
@michaelkirk I've pushed another commit PTAL

PR #215, opened 2017-05-12
use mic for voice message snippet

michaelkirk:
PTAL @charlesmchen


src/Messages/Attachments/TSAttachment.m
```diff
@@ -106,7 +106,16 @@ - (NSString *)description {
     } else if ([MIMETypeUtil isVideo:self.contentType]) {
         return [NSString stringWithFormat:@"📽 %@", attachmentString];
     } else if ([MIMETypeUtil isAudio:self.contentType]) {
-        return [NSString stringWithFormat:@"📻 %@", attachmentString];
+
+        // a missing filename is the legacy way to determin if an audio attachment is a voice note vs. other arbitrary
+        // audio attachments.
+        if (self.isVoiceMessage || !self.filename || self.filename.length == 0) {
```
charlesmchen, line 112: `self.filename.length` will be zero if self.filename is nil.

charlesmchen (approved):
LGTM

PR #216, opened 2017-05-12
Reduce time between editing contacts and seeing those changes in the app

michaelkirk:
* Move a couple lib methods behind our own interface
* Cache parsing phone numbers since it's expensive
* I considered caching formatting as well since it's also a bit
  expensive, but generating an appropriate cache key was actually slower
  than the raw implementation.

PTAL @charlesmchen

### DIFF
--- a/src/Contacts/PhoneNumberUtil.m
+++ b/src/Contacts/PhoneNumberUtil.m
@@ -52,6 +52,13 @@
 
     if (!result) {
         result = [self.nbPhoneNumberUtil parse:numberToParse defaultRegion:defaultRegion error:error];
+        if (error && *error) {
+            OWSAssert(!result);
+            return nil;
+        }
+
+        OWSAssert(result);
+
         if (result) {
             [self.parsedPhoneNumberCache setObject:result forKey:hashKey];
         } else {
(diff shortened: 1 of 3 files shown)


src/Contacts/PhoneNumberUtil.m
```diff
@@ -32,13 +34,45 @@ - (instancetype)init {
...
+                    defaultRegion:(NSString *)defaultRegion
+                            error:(NSError **)error
+{
+    NSString *hashKey = [NSString stringWithFormat:@"numberToParse:%@defaultRegion:%@", numberToParse, defaultRegion];
+
+    NBPhoneNumber *result = [self.parsedPhoneNumberCache objectForKey:hashKey];
+
+    if (!result) {
+        result = [self.nbPhoneNumberUtil parse:numberToParse defaultRegion:defaultRegion error:error];
+        if (result) {
```
charlesmchen: `if (result && !error) {`

and log error if non-nil?

charlesmchen: Also `error` parameter will be nil the second time we try to parse an invalid phone number, which might affect logging at caller. NBD?

charlesmchen (approved):
LGTM

src/Contacts/PhoneNumberUtil.m
michaelkirk: Good call, I opted to just never cache when there's an error.

PR #217, opened 2017-05-15
Charlesmchen/attachment filenames

charlesmchen:
* Rename attachment source filename property.
* Persist attachment file paths.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/TSAttachment.m
+++ b/src/Messages/Attachments/TSAttachment.m
@@ -88,7 +88,7 @@ NSUInteger const TSAttachmentSchemaVersion = 3;
     if (!_sourceFilename) {
         // renamed _filename to _sourceFilename
         _sourceFilename = [coder decodeObjectForKey:@"filename"];
-        OWSAssert(_sourceFilename || [_sourceFilename isKindOfClass:[NSString class]]);
+        OWSAssert(!_sourceFilename || [_sourceFilename isKindOfClass:[NSString class]]);
     }
 
     return self;
--- a/src/Messages/Attachments/TSAttachment.m
+++ b/src/Messages/Attachments/TSAttachment.m
@@ -22,7 +22,7 @@ NSUInteger const TSAttachmentSchemaVersion = 3;
 - (instancetype)initWithServerId:(UInt64)serverId
                    encryptionKey:(NSData *)encryptionKey
                      contentType:(NSString *)contentType
-                        filename:(nullable NSString *)filename
+                  sourceFilename:(nullable NSString *)sourceFilename
 {
     self = [super init];
     if (!self) {
@@ -33,14 +33,14 @@ NSUInteger const TSAttachmentSchemaVersion = 3;
     _encryptionKey = encryptionKey;
     _contentType = contentType;
     _attachmentSchemaVersion = TSAttachmentSchemaVersion;
-    _filename = filename;
+    _sourceFilename = sourceFilename;
 
     return self;
 }
 
 // This constructor is used for new instances of TSAttachmentStream
 // that represent new, un-uploaded outgoing attachments.
-- (instancetype)initWithContentType:(NSString *)contentType filename:(nullable NSString *)filename
+- (instancetype)initWithContentType:(NSString *)contentType sourceFilename:(nullable NSString *)sourceFilename
 {
     self = [super init];
     if (!self) {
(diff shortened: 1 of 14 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2116

src/Messages/Attachments/TSAttachment.m
```diff
@@ -85,6 +85,12 @@ - (nullable instancetype)initWithCoder:(NSCoder *)coder
         _attachmentSchemaVersion = TSAttachmentSchemaVersion;
     }
 
+    if (!_sourceFilename) {
+        // renamed _filename to _sourceFilename
+        _sourceFilename = [coder decodeObjectForKey:@"filename"];
+        OWSAssert(_sourceFilename || [_sourceFilename isKindOfClass:[NSString class]]);
```
michaelkirk: Won't this fail for some attachments e.g. voice notes from android

src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -112,30 +116,110 @@ + (NSUInteger)numberOfItemsInAttachmentsFolder
...
-                                      inFolder:[[self class] attachmentsFolder]];
+    if (!self.localRelativeFilePath) {
+        return nil;
+    }
+
+    return [[[self class] attachmentsFolder] stringByAppendingPathComponent:self.localRelativeFilePath];
+}
+
+- (nullable NSString *)localFilePathWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
```
michaelkirk: nit: This method is a little hairy - I imagine because we use it all over the place while the TSAttachment's are in a variety of states.

Also, I'm surprised that this method  might save the attachment. Why does it need to do that?

Would it simplify things to create the localfile path in the corresponding init methods?

michaelkirk (commented):
Is the motivation for this PR an insurance policy against some future code change that might cause us to lose reference to the files on disk?

charlesmchen:
> Is the motivation for this PR an insurance policy against some future code change that might cause us to lose reference to the files on disk?

Yes.

src/Messages/Attachments/TSAttachment.m
charlesmchen: Fixed a typo! Thanks.

src/Messages/Attachments/TSAttachmentStream.m
charlesmchen: Yes!  This approach is horrible... but I couldn't think of something better.

> Would it simplify things to create the localfile path in the corresponding init methods?

You're right. I was trying to avoid this since we don't want to create a transaction within a transaction, but this can be avoided by doing the write asynchronously.

charlesmchen:
PTAL @michaelkirk

charlesmchen:
PTAL @michaelkirk

src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -55,6 +57,20 @@ - (instancetype)initWithPointer:(TSAttachmentPointer *)pointer
     _isUploaded = YES;
     self.attachmentType = pointer.attachmentType;
 
+    [self ensureLocalFilePath:NO];
```
michaelkirk: nit: I prefer labeled parameters, even if it makes for verbose names.

src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -206,8 +206,12 @@ - (nullable NSURL *)mediaURL
 
 - (void)removeFileWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
 {
+    NSString *_Nullable localFilePath = self.localFilePath;
+    if (!localFilePath) {
+        return;
```
michaelkirk: Assertion failure // error log?

michaelkirk (commented):
LGTM

src/Messages/Attachments/TSAttachmentStream.m
charlesmchen: Done.

src/Messages/Attachments/TSAttachmentStream.m
charlesmchen: Done.

PR #218, opened 2017-05-15
remove checks for other country codes since it's expensive

michaelkirk:
Unfortunately calls to `NBPhoneNumberUtil parse:defaultRegion:error` are
not cheap, calling it with a bunch of permutations on every contact is
too expensive to justify it's incremental value unless we can get
upstream optimized.

e.g. for US numbers this was 26 extra calls per phone number

PTAL @charlesmchen 

**Before**

screen shot 2017-05-15 at 3 19 45 pm

**After**

screen shot 2017-05-15 at 4 16 20 pm

### DIFF
--- a/src/Contacts/PhoneNumber.m
+++ b/src/Contacts/PhoneNumber.m
@@ -169,18 +169,6 @@ static NSString *const RPDefaultsKeyPhoneNumberCanonical = @"RPDefaultsKeyPhoneN
                                        callingCodeForLocalNumber,
                                        sanitizedString],
                                       [self defaultRegionCode]);
-
-            // It's gratuitous to try all country codes associated with a given
-            // calling code, but it can't hurt and this isn't a performance
-            // hotspot.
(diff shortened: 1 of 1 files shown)


src/Contacts/PhoneNumber.m
```diff
@@ -169,18 +169,6 @@ + (PhoneNumber *)tryParsePhoneNumberFromUserSpecifiedText:(NSString *)text {
                                        callingCodeForLocalNumber,
                                        sanitizedString],
                                       [self defaultRegionCode]);
-
-            // It's gratuitous to try all country codes associated with a given
-            // calling code, but it can't hurt and this isn't a performance
-            // hotspot.
```
charlesmchen, line 175: > and this isn't a performance hotspot.

hmm...

charlesmchen (approved):
LGTM

src/Contacts/PhoneNumber.m
michaelkirk: 🤷‍♀️

PR #226, opened 2017-05-19
Pin YapDatabase to v2.9.3 to avoid v.3.x.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/SignalServiceKit.podspec
+++ b/SignalServiceKit.podspec
@@ -37,7 +37,7 @@ An Objective-C library for communicating with the Signal messaging service.
   s.dependency 'AFNetworking'
   s.dependency 'AxolotlKit'
   s.dependency 'Mantle'
-  s.dependency 'YapDatabase/SQLCipher'
+  s.dependency 'YapDatabase/SQLCipher', '~> 2.9.3'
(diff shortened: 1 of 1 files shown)


SignalServiceKit.podspec
```diff
@@ -37,7 +37,7 @@ An Objective-C library for communicating with the Signal messaging service.
   s.dependency 'AFNetworking'
   s.dependency 'AxolotlKit'
   s.dependency 'Mantle'
-  s.dependency 'YapDatabase/SQLCipher'
+  s.dependency 'YapDatabase/SQLCipher', '~> 2.9.3'
```
michaelkirk, line 40: LGTM, want to include the podfile.lock that you get after running pod install? It keeps future PR's more concise.

michaelkirk (approved):

SignalServiceKit.podspec
michaelkirk: oh, nevermind we're in SSK =P

charlesmchen: Yup, but that will be in Signal-iOS.

PR #227, opened 2017-05-20
Try the country code for the local phone number when parsing phone nu…

charlesmchen:
…mbers.

See: https://github.com/WhisperSystems/SignalServiceKit/pull/218/files

Unlike the code we reverted in #218, this will:

* Only try a single country code.
* Only try that country code if it differs from the device region code, so it most cases it won't have any effect.

PTAL @michaelkirk


charlesmchen:
PTAL @michaelkirk

michaelkirk (approved):
lgtm

PR #228, opened 2017-05-23
Cache image size and audio duration on attachments.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/Attachments/TSAttachmentStream.m
+++ b/src/Messages/Attachments/TSAttachmentStream.m
@@ -6,6 +6,7 @@
 #import "MIMETypeUtil.h"
 #import "TSAttachmentPointer.h"
 #import <AVFoundation/AVFoundation.h>
+#import <ImageIO/ImageIO.h>
 #import <YapDatabase/YapDatabase.h>
 #import <YapDatabase/YapDatabaseTransaction.h>
 
@@ -16,6 +17,11 @@ NS_ASSUME_NONNULL_BEGIN
 // We only want to generate the file path for this attachment once, so that
 // changes in the file path generation logic don't break existing attachments.
 @property (nullable, nonatomic) NSString *localRelativeFilePath;
+
+// These properties should only be accessed on the main thread.
+@property (nullable, nonatomic) NSNumber *cachedImageWidth;
+@property (nullable, nonatomic) NSNumber *cachedImageHeight;
+@property (nullable, nonatomic) NSNumber *cachedAudioDurationSeconds;
 
 @end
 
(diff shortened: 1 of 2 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2145

michaelkirk (approved):
LGTM

src/Messages/Attachments/TSAttachmentStream.m
```diff
@@ -289,6 +297,177 @@ + (void)deleteAttachments
...
+
+    return imageSize;
+}
+
+- (CGSize)cachedImageSizeWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    OWSAssert([NSThread isMainThread]);
+    OWSAssert(transaction);
+
+    if (self.cachedImageWidth && self.cachedImageHeight) {
```
michaelkirk, line 382: Isn't this redundant with the first few lines of `ensureCachedImageSizeWithTransaction`?

I also think the name of this method (cachedImageSizeWithTransaction) is confusing because you aren't necessarily getting a cached image size. I would expect a method named `cachedFoo` to return foo if it were cached, else return nil.

Can we just call `ensureCachedImageSizeWithTransaction` directly and delete this method? That would get rid of the redundancy and, I think be a better name.

charlesmchen: Sorry I just saw this comment.

charlesmchen: > Isn't this redundant with the first few lines of ensureCachedImageSizeWithTransaction?

Removed here and in the audio duration methods.

charlesmchen: > Can we just call ensureCachedImageSizeWithTransaction directly and delete this method? That would get rid of the redundancy and, I think be a better name.

I've renamed these methods by removing "cached":
```
- (CGSize)imageSizeWithTransaction:(YapDatabaseReadWriteTransaction *)transaction;
- (CGSize)imageSizeWithoutTransaction;

- (CGFloat)audioDurationSecondsWithTransaction:(YapDatabaseReadWriteTransaction *)transaction;
- (CGFloat)audioDurationSecondsWithoutTransaction;
```
Is that clearer?  

* I wanted to emphasize that a transaction may be used to avoid deadlock.
* If the user passes a transaction, we want it to be non-null.

WDYT?

See: https://github.com/WhisperSystems/Signal-iOS/pull/2149
See: https://github.com/WhisperSystems/SignalServiceKit/pull/229

PR #229, opened 2017-05-24
Rename audio duration and image size methods in TSAttachmentStream.

charlesmchen:
See: https://github.com/WhisperSystems/SignalServiceKit/pull/228
See: https://github.com/WhisperSystems/Signal-iOS/pull/2149

PTAL @michaelkirk


michaelkirk (approved):
LGTM, thanks!

PR #233, opened 2017-05-25
Charlesmchen/refine unseen indicator

charlesmchen:
Changes for unseen indicator.

* Add a database view for dynamic interactions.
* DRY up the creation of database views.
* Create separate database views for “unseen” and “unread” messages.
* Add “unseen tracking” to info and error messages.
* Rationalize “timestamp” vs. “receipt timestamp”.
* Ensure microsecond precision for interaction sorting.
* Add OWSFail() macros.

PTAL @michaelkirk

### DIFF
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -189,6 +189,23 @@ NS_ASSUME_NONNULL_BEGIN
     return hasUnread;
 }
 
+- (NSArray<id<OWSReadTracking>> *)unseenMessagesWithTransaction:(YapDatabaseReadTransaction *)transaction
+{
+    NSMutableArray<id<OWSReadTracking>> *messages = [NSMutableArray new];
+    [[transaction ext:TSUnseenDatabaseViewExtensionName]
+        enumerateRowsInGroup:self.uniqueId
+                  usingBlock:^(
+                      NSString *collection, NSString *key, id object, id metadata, NSUInteger index, BOOL *stop) {
+
+                      if (![object conformsToProtocol:@protocol(OWSReadTracking)]) {
+                          DDLogError(@"%@ Unexpected object in unseen messages: %@", self.tag, object);
+                      }
+                      [messages addObject:(id<OWSReadTracking>)object];
+                  }];
+
+    return [messages copy];
+}
+
 - (NSArray<id<OWSReadTracking> > *)unreadMessagesWithTransaction:(YapDatabaseReadTransaction *)transaction
 {
     NSMutableArray<id<OWSReadTracking> > *messages = [NSMutableArray new];
@@ -218,9 +235,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)markAllAsReadWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
 {
-    for (id<OWSReadTracking> message in [self unreadMessagesWithTransaction:transaction]) {
+    for (id<OWSReadTracking> message in [self unseenMessagesWithTransaction:transaction]) {
         [message markAsReadLocallyWithTransaction:transaction];
     }
+
+    // Just to be defensive, we'll also check for unread messages.
+    OWSAssert([self unreadMessagesWithTransaction:transaction].count < 1);
 }
 
 - (void)markAllAsRead
@@ -255,7 +275,7 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)updateWithLastMessage:(TSInteraction *)lastMessage transaction:(YapDatabaseReadWriteTransaction *)transaction {
-    NSDate *lastMessageDate = [lastMessage receiptDateForSorting];
+    NSDate *lastMessageDate = [lastMessage dateForSorting];
 
     if ([lastMessage isKindOfClass:[TSErrorMessage class]]) {
         TSErrorMessage *errorMessage = (TSErrorMessage *)lastMessage;
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -199,6 +199,8 @@ NS_ASSUME_NONNULL_BEGIN
 
                       if (![object conformsToProtocol:@protocol(OWSReadTracking)]) {
                           DDLogError(@"%@ Unexpected object in unseen messages: %@", self.tag, object);
+                          OWSFail(@"Unexpected object in unseen messages.");
+                          return;
                       }
                       [messages addObject:(id<OWSReadTracking>)object];
                   }];
--- a/src/Messages/Interactions/TSInfoMessage.m
+++ b/src/Messages/Interactions/TSInfoMessage.m
@@ -2,16 +2,40 @@
 //  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
 //
 
-#import "NSDate+millisecondTimeStamp.h"
 #import "TSInfoMessage.h"
+#import "NSDate+millisecondTimeStamp.h"
+#import <YapDatabase/YapDatabaseConnection.h>
 
 NS_ASSUME_NONNULL_BEGIN
+
+NSUInteger TSInfoMessageSchemaVersion = 1;
+
+@interface TSInfoMessage ()
+
+@property (nonatomic, getter=wasRead) BOOL read;
+
+@property (nonatomic, readonly) NSUInteger infoMessageSchemaVersion;
+
+@end
+
+#pragma mark -
 
 @implementation TSInfoMessage
 
 - (instancetype)initWithCoder:(NSCoder *)coder
 {
-    return [super initWithCoder:coder];
+    self = [super initWithCoder:coder];
+    if (!self) {
+        return self;
+    }
+
+    if (self.infoMessageSchemaVersion < 1) {
+        _read = YES;
+    }
+
+    _infoMessageSchemaVersion = TSInfoMessageSchemaVersion;
+
+    return self;
 }
 
 - (instancetype)initWithTimestamp:(uint64_t)timestamp
@@ -30,6 +54,7 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     _messageType = infoMessage;
+    _infoMessageSchemaVersion = TSInfoMessageSchemaVersion;
 
     return self;
 }
@@ -72,6 +97,40 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     return @"Unknown Info Message Type";
+}
+
+#pragma mark - OWSReadTracking
+
+- (BOOL)shouldAffectUnreadCounts
+{
+    return NO;
+}
+
+- (void)markAsReadLocally
+{
+    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+        [self markAsReadLocallyWithTransaction:transaction];
+    }];
+}
+
+- (void)markAsReadLocallyWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
+{
+    OWSAssert(transaction);
+    DDLogInfo(@"%@ marking as read uniqueId: %@ which has timestamp: %llu", self.tag, self.uniqueId, self.timestamp);
+    _read = YES;
+    [self saveWithTransaction:transaction];
+}
+
+#pragma mark - Logging
+
++ (NSString *)tag
+{
+    return [NSString stringWithFormat:@"[%@]", self.class];
+}
+
+- (NSString *)tag
+{
+    return self.class.tag;
 }
 
 @end
--- a/src/Messages/Interactions/TSInteraction.h
+++ b/src/Messages/Interactions/TSInteraction.h
@@ -16,7 +16,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) TSThread *thread;
 @property (nonatomic, readonly) uint64_t timestamp;
 
-- (NSDate *)date;
 - (NSString *)description;
 
 /**
@@ -33,7 +32,10 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)interactionForTimestamp:(uint64_t)timestamp
                         withTransaction:(YapDatabaseReadWriteTransaction *)transaction;
 
-- (nullable NSDate *)receiptDateForSorting;
+// NSDate has second precision, uint64_t timestamps have millisecond precision
+// so prefer timestampForSorting over dateForSorting.
+- (NSDate *)dateForSorting;
+- (uint64_t)timestampForSorting;
 
 // "Dynamic" interactions are not messages or static events (like
 // info messages, error messages, etc.).  They are interactions
--- a/src/Messages/Interactions/TSInteraction.h
+++ b/src/Messages/Interactions/TSInteraction.h
@@ -32,10 +32,9 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)interactionForTimestamp:(uint64_t)timestamp
                         withTransaction:(YapDatabaseReadWriteTransaction *)transaction;
 
-// NSDate has second precision, uint64_t timestamps have millisecond precision
-// so prefer timestampForSorting over dateForSorting.
 - (NSDate *)dateForSorting;
 - (uint64_t)timestampForSorting;
+- (NSComparisonResult)compareForSorting:(TSInteraction *)other;
 
 // "Dynamic" interactions are not messages or static events (like
 // info messages, error messages, etc.).  They are interactions
--- a/src/Messages/Interactions/TSInteraction.h
+++ b/src/Messages/Interactions/TSInteraction.h
@@ -35,6 +35,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (nullable NSDate *)receiptDateForSorting;
 
+// "Dynamic" interactions are not messages or static events (like
+// info messages, error messages, etc.).  They are interactions
+// created, updated and deleted by the views.
+//
+// These include block offers, "add to contact" offers,
+// unseen message indicators, etc.
+- (BOOL)isDynamicInteraction;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/src/Messages/Interactions/TSInteraction.m
+++ b/src/Messages/Interactions/TSInteraction.m
@@ -109,6 +109,11 @@ NS_ASSUME_NONNULL_BEGIN
     [fetchedThread updateWithLastMessage:self transaction:transaction];
 }
 
+- (BOOL)isDynamicInteraction
+{
+    return NO;
+}
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/src/Messages/Interactions/TSInteraction.m
+++ b/src/Messages/Interactions/TSInteraction.m
@@ -3,6 +3,7 @@
 //
 
 #import "TSInteraction.h"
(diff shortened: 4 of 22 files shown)


src/Messages/Interactions/TSInteraction.h
```diff
@@ -33,7 +32,18 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)interactionForTimestamp:(uint64_t)timestamp
                         withTransaction:(YapDatabaseReadWriteTransaction *)transaction;
 
-- (nullable NSDate *)receiptDateForSorting;
+// NSDate has second precision, uint64_t timestamps have millisecond precision
+// so prefer timestampForSorting over dateForSorting.
+- (NSDate *)dateForSorting;
```
michaelkirk: Though the *units* are seconds, they are stored as doubles, so NSDate already has submillisecond precision that we can sort with. 

e.g. 
```
po [[NSDate date] timeIntervalSince1970];
1495751430.6830649
```

You fixed a good example where we were not preserving that precision when building the NSDate. Now that it's fixed, do you still want to make all the changes related to the addition of the `timestampForSorting`?

src/Messages/Interactions/TSInteraction.m
```diff
@@ -72,11 +72,6 @@ - (uint64_t)millisecondsTimestamp {
     return self.timestamp;
 }
 
-- (NSDate *)date {
-    uint64_t seconds = self.timestamp / 1000;
```
michaelkirk, line 76: We were losing that subsecond precision here, good catch!

src/Messages/OWSUnknownContactBlockOfferMessage.m
```diff
@@ -34,14 +34,16 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp thread:(TSThread *)thread
...
-    //
-    // By default [TSMessage receiptDateForSorting] will prefer to use receivedAtDate
-    // which is not back-dated.
-    return self.date;
+    // Use the timestamp, not the "received at" timestamp to sort,
+    // since we're creating these interactions after the fact and back-dating them.
+    return NO;
+}
+
+- (BOOL)isDynamicInteraction
```
michaelkirk, line 44: take it or leave it: 

Is it true that `shouldUseRecieptDataForSorting = YES` iff `isDynamicInteraction = NO`

If that's true, maybe `shouldUseReceiptDateForSorting` should just be `!isDynamicInteraction`

Unless I'm missing a case.

src/Storage/TSDatabaseView.h
```diff
@@ -19,11 +15,21 @@ extern NSString *TSSecondaryDevicesGroup;
 extern NSString *TSThreadDatabaseViewExtensionName;
 extern NSString *TSMessageDatabaseViewExtensionName;
 extern NSString *TSUnreadDatabaseViewExtensionName;
+extern NSString *TSUnseenDatabaseViewExtensionName;
+extern NSString *TSDynamicMessagesDatabaseViewExtensionName;
 extern NSString *TSSecondaryDevicesDatabaseViewExtensionName;
 
 + (BOOL)registerThreadDatabaseView;
-+ (BOOL)registerBuddyConversationDatabaseView;
++ (BOOL)registerThreadInteractionsDatabaseView;
```
michaelkirk, line 24: This may be a clearer name, but it lacks that kool 90's vibe.

src/Storage/TSDatabaseView.h
```diff
@@ -19,11 +15,21 @@ extern NSString *TSSecondaryDevicesGroup;
...
+// Instances of OWSReadTracking for wasRead is NO and shouldAffectUnreadCounts is YES.
+//
+// Should be used for "unread message counts".
 + (BOOL)registerUnreadDatabaseView;
+// Should be used for "unread indicator".
```
michaelkirk: nit: empty space above block comments.

src/Storage/TSDatabaseView.m
```diff
@@ -183,19 +218,12 @@ + (YapDatabaseViewSorting *)messagesSorting {
           TSInteraction *message1 = (TSInteraction *)object1;
           TSInteraction *message2 = (TSInteraction *)object2;
 
-          NSDate *date1 = [self localTimeReceiveDateForInteraction:message1];
-          NSDate *date2 = [self localTimeReceiveDateForInteraction:message2];
+          uint64_t timestamp1 = message1.timestampForSorting;
+          uint64_t timestamp2 = message2.timestampForSorting;
 
-          NSComparisonResult result = [date1 compare:date2];
```
michaelkirk: It seems like this code (and lots of pre-existing code) was written on the premise that NSDate is seconds granularity.

But I think the whole method could be replaced with:
```
return [message1.dateForSorting compare:message2.dateForSorting]
```

Since NSDate is double precision.

src/Storage/TSStorageManager.m
```diff
@@ -197,8 +197,10 @@ - (void)setupDatabase
 {
     // Register extensions which are essential for rendering threads synchronously
     [TSDatabaseView registerThreadDatabaseView];
-    [TSDatabaseView registerBuddyConversationDatabaseView];
+    [TSDatabaseView registerThreadInteractionsDatabaseView];
```
michaelkirk, line 200: Reminder to test this migration on a really slow device with a lot of messages.

src/Util/Asserts.h
```diff
@@ -13,24 +13,40 @@
...
+        NSAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                               \
+    }
+
+#define OWSCAssert(X)                                                                                                  \
+    if (!(X)) {                                                                                                        \
+        NSLog(@"%s Assertion failed: %s", __PRETTY_FUNCTION__, CONVERT_EXPR_TO_STRING(X));                             \
+        NSCAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                              \
+    }
+
+#define OWSFail(X)                                                                                                     \
```
michaelkirk, line 31: Awesome. Thanks!

On a more radical note, what if we defined these in non-debug, and changed the NSLog to DDLogError so that we can see them in production logs (and eventually they could be reported).

No need to do that in this PR, if ever.

src/Contacts/TSThread.m
```diff
@@ -183,6 +183,23 @@ - (BOOL)hasUnreadMessages {
...
+- (NSArray<id<OWSReadTracking>> *)unseenMessagesWithTransaction:(YapDatabaseReadTransaction *)transaction
+{
+    NSMutableArray<id<OWSReadTracking>> *messages = [NSMutableArray new];
+    [[transaction ext:TSUnseenDatabaseViewExtensionName]
+        enumerateRowsInGroup:self.uniqueId
+                  usingBlock:^(
+                      NSString *collection, NSString *key, id object, id metadata, NSUInteger index, BOOL *stop) {
+
+                      if (![object conformsToProtocol:@protocol(OWSReadTracking)]) {
+                          DDLogError(@"%@ Unexpected object in unseen messages: %@", self.tag, object);
```
michaelkirk, line 201: assert and return

src/Messages/Interactions/TSInfoMessage.m
```diff
@@ -2,16 +2,40 @@
...
 //
 
-#import "NSDate+millisecondTimeStamp.h"
 #import "TSInfoMessage.h"
+#import "NSDate+millisecondTimeStamp.h"
+#import <YapDatabase/YapDatabaseConnection.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSUInteger TSInfoMessageSchemaVersion = 1;
```
michaelkirk, line 11: Just for discussion, it seems like there's no real reason to have separate classes for error vs. info message in the DB, right? It feels like the distinction only exists so as to print a red vs yellow border, which we could achieve much more simply. (not for this PR)

src/Messages/Interactions/TSMessage.m
```diff
@@ -104,7 +111,12 @@ - (instancetype)initWithTimestamp:(uint64_t)timestamp
     _expiresInSeconds = expiresInSeconds;
     _expireStartedAt = expireStartedAt;
     [self updateExpiresAt];
-    _receivedAtDate = [NSDate date];
+
+    if ([self shouldUseReceiptDateForSorting]) {
```
michaelkirk: These conditions read backwards. Isn't "now" (aka `[NSDate ows_millisecondTimeStamp];` or `[NSDate new]` the receipt date?

src/Storage/TSDatabaseView.m
```diff
@@ -22,40 +20,106 @@
...
                 return possiblyRead.uniqueThreadId;
             }
         }
         return nil;
     }];
 
-    YapDatabaseViewSorting *viewSorting = [self messagesSorting];
+    return [self registerMessageDatabaseViewWithName:TSUnreadDatabaseViewExtensionName
+                                        viewGrouping:viewGrouping
+                                             version:@"1"];
```
michaelkirk, line 71: We've changed the logic of this view, but since the only things that newly have a "read" property (info/error messages) also are already migrated as `read=YES` and they `!shouldAffectUnreadCounts` anyway so... I guess we don't have to bump the version to have clients rebuild their index since it should result in the same output.

...right?

src/Storage/TSDatabaseView.m
```diff
@@ -22,40 +20,106 @@
...
+}
+
++ (BOOL)registerUnreadDatabaseView
+{
     YapDatabaseViewGrouping *viewGrouping = [YapDatabaseViewGrouping withObjectBlock:^NSString *(
         YapDatabaseReadTransaction *transaction, NSString *collection, NSString *key, id object) {
         if ([object conformsToProtocol:@protocol(OWSReadTracking)]) {
             id<OWSReadTracking> possiblyRead = (id<OWSReadTracking>)object;
-            if (possiblyRead.read == NO) {
+            if (!possiblyRead.wasRead && possiblyRead.shouldAffectUnreadCounts) {
```
michaelkirk, line 62: It seems like our old concept of "unread" we now refer to as "unseen" and the new concept of "unread" is `unseen && shouldAffectUnreadCount`

So, should we rename the field `wasRead` to `wasSeen`?

Would that be clearer, and if so, is it worth the change - it affects several models? WDYT?

src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyErrorMessage.m
```diff
@@ -29,6 +25,11 @@ - (NSString *)theirSignalId
     return nil;
 }
 
+- (BOOL)isDynamicInteraction
+{
+    return YES;
```
michaelkirk: I haven't really seen how this is used yet - presumably it'll come up in a forthcoming Signal-iOS pull request, but Identity key errors seem very different from contact offers / blocking offers.

My immediate concern is that the others are treated as sort of disposable because they can be recreated, but the identity key error message can not. What benefit/functionality do we get by marking them as dynamic?

michaelkirk (commented):
The overall approach seems solid, but I'd like to re-review once I see how this is integrated into Signal-iOS

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2157

@michaelkirk PTAL

src/Messages/Interactions/TSInteraction.m
charlesmchen: Hmm, this problem exists elsewhere too.  I'm pushing another commit that fixes it.

src/Messages/OWSUnknownContactBlockOfferMessage.m
charlesmchen: I thought about it. It's true today, but it seemed like the kind of thing that could introduce a very-hard to debug issue if either of us forgot or moved on.

src/Storage/TSDatabaseView.h
charlesmchen: Done.

src/Storage/TSDatabaseView.m
charlesmchen: Yes, these were tough choices, and I'm totally open to going through them carefully.

> It seems like our old concept of "unread" we now refer to as "unseen" and the new concept of "unread" is unseen && shouldAffectUnreadCount

* I would put it differently.  I'd say that we now have concepts of "unread that affects unread counts" and "unread".  
* I've mostly avoided using the term "unseen" in our codebase because: a) we're already using "unread" all over the place. b) desire for consistency. c) we use "unread" not "unseen" in our user-facing copy. 

> So, should we rename the field wasRead to wasSeen?

We could - but why go to the trouble?  In particular, you'll note that I landed upon using a single piece of state, since there is no scenario in which they have different values.

Or, let me ask - how would that be clearer?

src/Storage/TSDatabaseView.m
charlesmchen: That's right. One of the nice things about leaving "unread" as "unread" is that it greatly reduces the amount of changes like this.  This index should operate exactly as before - on the same set of objects, tracking the same values of the same properties which still have the same names.

src/Messages/Interactions/TSInteraction.h
charlesmchen: Great point. 

> Now that it's fixed, do you still want to make all the changes related to the addition of the timestampForSorting?

I think so.  Why?

* We're already heavily committed to using `uint64_t` for `timestamp`.  It's nice to migrate `receivedAtTimestamp` to the same date type for consistency.
* We're carefully constructing the timestamps of the dynamic interactions to control the ordering of interactions in message view (See: `kBlockOfferOffset`, `kAddToContactsOfferOffset`, `kUnreadIndicatorOfferOffset`, etc.).  Here too it's nice to use a consistent datatype.
* I've just added a `compareForSorting` method to TSInteraction that dries up a lot of sorting logic.

charlesmchen: > carefully constructing the timestamps 

This is the key argument.  It's nice to use really small `theta` (aka increment) values to order the interactions, and it's important that we don't have to worry about whether values are getting ceil()'d, round()'d or floor()'d as they move back and forth between `NSDate` and `uint64_t`.

src/Storage/TSDatabaseView.m
charlesmchen: See above!  Also, I've dried up all of this comparison code with `compareForSorting:`.

src/Storage/TSStorageManager.m
charlesmchen: Sorry, which migration?  That's a really key scenario to test with this work (I've already been testing it a ton), but this particular database view isn't changing - I've just renamed it in the code, but not changed its extension name.

src/Util/Asserts.h
charlesmchen: > On a more radical note, what if we defined these in non-debug, and changed the NSLog to DDLogError so that we can see them in production logs (and eventually they could be reported).

I'd love to do that logging in production - but not actually use the asserts.  I think it's key for stability that our code fails gracefully in production.  I'm pretty sure we're rolling past asserts like crazy in production right now. 

And the NSLog usage here isn't great. But I haven't converted them to DDLogError() because I was worried about PII leaking into the production logs.  

My long-term hope is to unify:

* Logging
* Asserts
* Analytics
* Return values from these macros so that we can be more explicit about how to handle deviations from these invariants in production. 

I'm hoping that we could land on a _small_ handful of macros like:

```
// This macro would DDLogError(), Assert in DEBUG, send an analytics event, return a value that indicates the assert failed... all in a tidy little consistent bit of code.
if (!DDCriticalAssertWAnalytics(message != nil, @"Incoming message failed to decrypt")) {
   return;
}
```

WDYT?

charlesmchen: Oh and since we're gabbing, I also intend to modify these macros to take NSString-style  format + varargs arguments.  But I haven't done it yet.

src/Contacts/TSThread.m
charlesmchen: Done.

src/Messages/Interactions/TSInfoMessage.m
charlesmchen: > It feels like the distinction only exists so as to print a red vs yellow border, which we could achieve much more simply.

Strongly agree!  There's a lot of classes that could be combined.  This would also facilitate our doing something like "make safety number changes switch from red "error" border to orange "info" border after user has visited verification view for that change.

src/Messages/Interactions/TSMessage.m
charlesmchen: Thank you - I got this wrong.

src/Messages/InvalidKeyMessages/TSInvalidIdentityKeyErrorMessage.m
charlesmchen: I was wrestling with whether or not to do this.  My thoughts were:

* It's possible these will become dynamic - we might delete them to de-duplicate them, we might change them to reflect whether or not they have been verified - but it's true they are not dynamic yet.
* These are a lot like the dynamic interactions - rare but of great interest.
* Database views and queries aren't free and I foresaw us often querying for the dynamic interactions and safety number changes at the same time and in the same way. 

So it was a bit of "let's be pragmatic and not puritanical"... but you've talked me out of it.  I've added another database view.

charlesmchen:
@michaelkirk PTAL I pushed a commit responding to your comments.

src/Storage/TSStorageManager.m
michaelkirk: Sorry, sloppy comment, let me try again. =)

We're making a couple new DB views. Each of these will iterate over every object in the data store, which can take some time.

During QA, we should verify the upgrade process is reasonably smooth for slow devices that have huge message collections.

src/Util/Asserts.h
michaelkirk: > WDYT?
 
I think we'll need to workshop the name. =P

But otherwise sounds right on.

src/Storage/TSStorageManager.m
charlesmchen: Gotcha.

charlesmchen:
@michaelkirk PTAL I pushed another commit responding to your comments.

src/Messages/OWSUnknownContactBlockOfferMessage.m
michaelkirk: fair enough!

src/Storage/TSDatabaseView.m
michaelkirk: well... it seems a little weird that we have a set of 'unseen messages", and inclusion in the set is determined by !wasRead. 

And we have another set of "unread messages", and inclusion in the set is determined by "!wasUnread && shouldAffectUnreadCount"

It's fine so long as you don't read the words.

michaelkirk: I'm half joking - I do think it's a little unclear, but the distinction is pretty small and the ramifications of cleaning it up seem pretty heavy relative to the value. I'd say leave it as is.

src/Messages/Interactions/TSInteraction.h
michaelkirk: I'm OK with this, and further, it's clearly easy to forget about precision when dealing in seconds units, given the number of bugs you fixed around this. I think dealing with the raw milliseconds makes this a little more me-proof.

michaelkirk (approved):
LGTM!

PR #235, opened 2017-05-30
Fix “mark as read” logic.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -242,14 +242,14 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     // Just to be defensive, we'll also check for unread messages.
-    OWSAssert([self unreadMessagesWithTransaction:transaction].count < 1);
+    OWSAssert([self unseenMessagesWithTransaction:transaction].count < 1);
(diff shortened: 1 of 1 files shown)


src/Contacts/TSThread.m
```diff
@@ -242,14 +242,14 @@ - (void)markAllAsReadWithTransaction:(YapDatabaseReadWriteTransaction *)transact
     }
 
     // Just to be defensive, we'll also check for unread messages.
-    OWSAssert([self unreadMessagesWithTransaction:transaction].count < 1);
+    OWSAssert([self unseenMessagesWithTransaction:transaction].count < 1);
 }
 
 - (void)markAllAsRead
 {
-    for (id<OWSReadTracking> message in [self unreadMessages]) {
```
michaelkirk, line 250: Is `unreadMessages` method still used? Can it go away now?

michaelkirk (approved):
LGTM

src/Contacts/TSThread.m
charlesmchen: Removed.

PR #239, opened 2017-05-31
dedicated session connection

michaelkirk:
Originally we talked about doing this a little later, but I wanted to get it in this release since there's potentially new ways a mutated session could go unsaved (more places we're throwing exceptions) with the 2.13 release.

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #241, opened 2017-05-31
determine if recipient identity change is unseen

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/src/Storage/AxolotlStore/TSStorageManager+IdentityKeyStore.m
+++ b/src/Storage/AxolotlStore/TSStorageManager+IdentityKeyStore.m
@@ -278,6 +278,24 @@ const NSTimeInterval kIdentityKeyStoreNonBlockingSecondsThreshold = 5.0;
     }
 }
 
+
+- (BOOL)hasUnseenIdentityChangeForRecipientId:(NSString *)recipientId
+{
+    OWSAssert(recipientId != nil);
+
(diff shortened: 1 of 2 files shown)


src/Storage/AxolotlStore/TSStorageManager+IdentityKeyStore.m
```diff
@@ -199,6 +199,31 @@ - (nullable OWSRecipientIdentity *)unconfirmedIdentityThatShouldBlockSendingForR
     }
 }
 
+- (nullable OWSRecipientIdentity *)unseenIdentityChangeForRecipientId:(NSString *)recipientId
+{
+    OWSAssert(recipientId != nil);
+
+    @synchronized([[self class] sharedIdentityKeyLock])
+    {
+        OWSRecipientIdentity *currentIdentity = [OWSRecipientIdentity fetchObjectWithUniqueID:recipientId];
```
charlesmchen, line 208: Should we use `sessionDBConnection` for identities?

charlesmchen (approved):
LGTM

src/Storage/AxolotlStore/TSStorageManager+IdentityKeyStore.m
michaelkirk: Hmmm... why? Keep in mind that OWSRecipientIdentity already have a similarly constrained (no-cache) dbConnection.

charlesmchen: Should we combine these two connections?  NBD.

PR #244, opened 2017-06-05
Mkirk/fix tests

michaelkirk:
PTAL @charlesmchen 

I think it's all non-controversial, except maybe:
https://github.com/WhisperSystems/SignalServiceKit/commit/819f3f1eb98910625ae15a420e3009bf5444bc9c


src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -121,7 +121,7 @@ - (void)runLoop
 
 + (void)setExpirationForMessage:(TSMessage *)message
 {
-    dispatch_async(self.serialQueue, ^{
+    dispatch_sync(self.serialQueue, ^{
```
charlesmchen: I don't think we want to do this. This queue can get a bit backed up. If we need to make this testable, we could just have the test wait, or add a completion, etc.

Example/TSKitiOSTestApp/Podfile
```diff
@@ -11,3 +11,18 @@ target 'TSKitiOSTestApp' do
...
+  installer.pods_project.targets.each do |target|
+    if target.to_s == "SignalServiceKit"
+      puts "--[!] Disabling singleton enforcement for SSK."
+      target.build_configurations.each do |config|
+        existing_definitions = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS']
+        if existing_definitions == nil || existing.length == 0
+          existing_definitions = "$(inheritied)"
+        end
+
+        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] = "#{existing_definitions} SSK_NO_ENFORCE_SINGLETONS=1"
```
charlesmchen: Instead of SSK_NO_ENFORCE_SINGLETONS how about BUILDING_FOR_TESTS or something generic like that?

michaelkirk: SSK_BUILDING_FOR_TESTS works for me.

src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: @michaelkirk Did you see this comment?

michaelkirk: Yup, that's why it's not merged. 

I think I'd prefer to add a completion handler vs. getting in the habit of adding sleep(1) in tests, but that would take more than a minute of dev time. 

This PR hasn't been top priority for me, so feel free to run with it if you're feeling frisky.

michaelkirk: PTAL @charlesmchen 

I considered adding a completion handler, but the method chain is really deep, across multiple dispatch_asyncs. So I opted for a `sleep` hack in the tests. I really don't like to do this, but I think it's more important we get the tests running.

e.g. since I last picked up this pull request there were 3 or so other changes that broke a bunch of tests. Commits to fix those are also included in this PR.

charlesmchen (approved):
LGTM

PR #245, opened 2017-06-05
Rework and unify the system messages.

charlesmchen:
PTAL @michaelkirk


michaelkirk (approved):
LGTM

charlesmchen:
@michaelkirk It's possible I was too overzealous with these asserts; we may need to revert part of this PR.

PR #246, opened 2017-06-05
Charlesmchen/format fail messages

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Util/Asserts.h
+++ b/src/Util/Asserts.h
@@ -28,25 +28,27 @@
         NSCAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                              \
     }
 
-#define OWSFail(X)                                                                                                     \
+#define OWSFail(message, ...)                                                                                          \
     {                                                                                                                  \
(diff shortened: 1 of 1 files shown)


michaelkirk (approved):
Great!

LGTM

src/Util/Asserts.h
```diff
@@ -28,25 +28,27 @@
         NSCAssert(0, @"Assertion failed: %s", CONVERT_EXPR_TO_STRING(X));                                              \
     }
 
-#define OWSFail(X)                                                                                                     \
+#define OWSFail(message, ...)                                                                                          \
```
michaelkirk: Do we need to use NS_FORMAT_FUNCTION to ensure type of the passed params?

charlesmchen: I don't think so. I think NS_FORMAT_FUNCTION is used to decorate the declarations of  functions and methods. It's not for macros.  For example, in this case, the macro will be preprocessed to `[NSString stringWithFormat:...]` whose declaration has `NS_FORMAT_FUNCTION`: 
```
+ (instancetype)stringWithFormat:(NSString *)format, ... NS_FORMAT_FUNCTION(1,2);
```

PR #252, opened 2017-06-07
Charlesmchen/sync verification state

charlesmchen:
PTAL @michaelkirk


protobuf/OWSSignalServiceProtos.proto
```diff
@@ -124,12 +124,26 @@ message SyncMessage {
...
-  optional Groups   groups   = 3;
-  optional Request  request  = 4;
-  repeated Read     read     = 5;
-  optional Blocked  blocked  = 6;
+  message Verification {
+    enum State {
+      DEFAULT            = 0;
+      VERIFIED           = 1;
+      NO_LONGER_VERIFIED = 2;
+    }
```
charlesmchen, line 132: These changes are temporary until Moxie's proto schema changes are merged.

src/Devices/OWSVerificationStateSyncMessage.h
```diff
@@ -0,0 +1,21 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSVerificationStateSyncMessage : OWSOutgoingSyncMessage
+
+- (void)addVerificationState:(OWSVerificationState)verificationState
+                 identityKey:(NSData *)identityKey
+                 recipientId:(NSString *)recipientId;
+
+// Returns the list of recipient ids referenced in this message.
+- (NSArray<NSString *> *)recipientIds;
```
michaelkirk, line 17: nit: readonly property for idempotent getters?

src/Devices/OWSVerificationStateSyncMessage.m
```diff
@@ -0,0 +1,102 @@
...
+
+- (void)addVerificationState:(OWSVerificationState)verificationState
+                 identityKey:(NSData *)identityKey
+                 recipientId:(NSString *)recipientId
+{
+    OWSAssert(identityKey.length > 0);
+    OWSAssert(recipientId.length > 0);
+    OWSAssert(self.tuples);
+
+    OWSVerificationStateTuple *tuple = [OWSVerificationStateTuple new];
```
michaelkirk, line 56: take it or leave it: Did you consider using the existing OWSRecipientIdentity instead of introducing this new class with a subset of fields? You could do a `copy` if you're worried about it mutating.

protobuf/OWSSignalServiceProtos.proto
```diff
@@ -124,12 +124,26 @@ message SyncMessage {
...
+
+    optional State  state       = 1;
+    // The e164 phone number of the user.
+    optional string destination = 2;
```
michaelkirk, line 136: I think `recipientId` would be a better name. Destination is overloaded to mean "where a message is headed".

src/Devices/OWSVerificationStateSyncMessage.m
```diff
@@ -0,0 +1,102 @@
...
+}
+
+- (OWSSignalServiceProtosSyncMessage *)buildSyncMessage
+{
+    OWSAssert(self.tuples.count > 0);
+    
+    OWSSignalServiceProtosSyncMessageBuilder *syncMessageBuilder = [OWSSignalServiceProtosSyncMessageBuilder new];
+    for (OWSVerificationStateTuple *tuple in self.tuples) {
+        OWSSignalServiceProtosSyncMessageVerificationBuilder *verificationBuilder = [OWSSignalServiceProtosSyncMessageVerificationBuilder new];
+        [verificationBuilder setDestination:tuple.recipientId];
```
michaelkirk: nit: you can use regular ole' property syntax with protobufs: `verificationBuilder.destination = tuple.recipientid`

src/Devices/OWSVerificationStateSyncMessage.m
```diff
@@ -0,0 +1,102 @@
...
+
+- (NSArray<NSString *> *)recipientIds
+{
+    NSMutableArray<NSString *> *result = [NSMutableArray new];
+    for (OWSVerificationStateTuple *tuple in self.tuples) {
+        OWSAssert(tuple.recipientId.length > 0);
+        [result addObject:tuple.recipientId];
+    }
+    
+    return result;
```
michaelkirk: nit: copy to return immutable instance

src/Messages/OWSBlockingManager.m
```diff
@@ -201,6 +214,15 @@ - (void)lazyLoadBlockedPhoneNumbersIfNecessary
                          inCollection:kOWSBlockingManager_BlockedPhoneNumbersCollection];
     _blockedPhoneNumberSet = [[NSMutableSet alloc] initWithArray:(blockedPhoneNumbers ?: [NSArray new])];
 
+    [self syncBlockedPhoneNumbersIfNecessary];
```
michaelkirk, line 217: nit: maybe this method name (`lazyLoadBlockedPhoneNumbersIfNecessary`) deserves a promotion since it's getting more responsibilities. Like `vicePresidentOfEnsuringSetup` or something.

src/Messages/OWSBlockingManager.m
```diff
@@ -201,6 +214,15 @@ - (void)lazyLoadBlockedPhoneNumbersIfNecessary
...
+    [self observeNotifications];
+}
+
+// This method should only be called from within a synchronized block.
```
michaelkirk, line 221: > +// This method should only be called from within a synchronized block.

Do we need to `synchronize(self)` the call from `applicationDidBecomeActive`?

michaelkirk: And can we make `blockedPhoneNumberSet` an atomic property since we're potentially reading from it on different threads, and sometimes re-assigning it.

src/Messages/OWSIdentityManager.m
```diff
@@ -186,24 +214,33 @@ - (void)setVerificationState:(OWSVerificationState)verificationState
...
-        [identity updateWithVerificationState:verificationState];
+        [recipientIdentity updateWithVerificationState:verificationState];
+
+        if (sendSyncMessage) {
+            [self enqueueSyncMessageForVerificationState:verificationState
+                                             identityKey:identityKey
+                                             recipientId:recipientId];
+        } else {
+            // Cancel any pending verification state sync messages for this recipient.
+            [self clearSyncMessageForRecipientId:recipientId];
```
michaelkirk, line 242: Just typing this out to make sure I understand...

...so **receiving** a verification sync will cancel an in-progress sending sync. 

Presumably this is most relevant when bringing a device back online, after being offline for a while, in which case any locally queued outgoing verification sync message is probably less relevant than any that have happened since on the secondary device.

So, I *think* this will usually be right. I don't think we *can* be right in every case, which is why we have an audit trail of info messages. 👍

src/Messages/OWSIdentityManager.m
```diff
@@ -361,6 +398,286 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
+                                           inCollection:OWSIdentityManager_QueuedVerificationStateSyncMessages];
+            }
+        }
+    });
+}
+
+- (void)processIncomingSyncMessage:(NSArray<OWSSignalServiceProtosSyncMessageVerification *> *)verifications
+{
+    for (OWSSignalServiceProtosSyncMessageVerification *verification in verifications) {
+        NSString *recipientId = [verification destination];
```
michaelkirk: nit: I think regular property getters are generally more readable. Sometimes I use the message-send style when I'm trying to convey something as more expensive than a simple "get" but... that's probably just a silly thing I do.

src/Messages/OWSIdentityManager.m
```diff
@@ -361,6 +398,286 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
+                                             overwriteOnConflict:NO];
+                break;
+            case OWSSignalServiceProtosSyncMessageVerificationStateVerified:
+                [self tryToApplyVerificationStateFromSyncMessage:OWSVerificationStateVerified
+                                                     recipientId:recipientId
+                                                     identityKey:identityKey
+                                             overwriteOnConflict:YES];
+                break;
+            case OWSSignalServiceProtosSyncMessageVerificationStateNoLongerVerified:
+                OWSFail(@"Verification state sync message for recipientId: %@ has unexpected value: %@.",
```
michaelkirk, line 561: nit: Can we log this?

src/Messages/OWSIdentityManager.m
```diff
@@ -361,6 +398,286 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
+                return;
+            }
+
+            // Ensure a remote identity exists for this key. We may be learning about
+            // it for the first time.
+            [self saveRemoteIdentity:identityKey recipientId:recipientId];
+
+            recipientIdentity = [OWSRecipientIdentity fetchObjectWithUniqueID:recipientId];
+
+            if (recipientIdentity == nil) {
```
michaelkirk, line 602: can we error log all these OWSFail? Or use a macro that fails and DDLogs?

src/Messages/OWSIdentityManager.m
```diff
@@ -361,6 +398,286 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
+            if (![recipientIdentity.identityKey isEqualToData:identityKey]) {
+                OWSFail(@"recipientIdentity has unexpected identityKey");
+                return;
+            }
+
+            if (recipientIdentity.verificationState == verificationState) {
+                return;
+            }
+
+            DDLogInfo(@"%@ setVerificationState: %@ (%@ -> %@)",
```
michaelkirk, line 621: maybe clearer: **sync**VerificationState: %@ (%@ -> %@), but then IIRC this is also logged in `updateWithVerificationState` so maybe this can just go away.

Feel free to ignore this comment =)

src/Messages/OWSIdentityManager.m
```diff
@@ -361,6 +398,286 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
+                OWSFail(@"recipientIdentity has unexpected recipientId: %@", recipientId);
+                return;
+            }
+
+            if (![recipientIdentity.identityKey isEqualToData:identityKey]) {
+                // The conflict case where we receive a verification sync message
+                // whose identity key disagrees with the local identity key for
+                // this recipient.
+                if (!overwriteOnConflict) {
+                    DDLogWarn(@"recipientIdentity has non-matching identityKey");
```
michaelkirk: nit: can you include the recipient ID in all these log messages?

michaelkirk (approved):
Other than the concurrency concern commented on in the blocking manager, this LGTM

src/Devices/OWSVerificationStateSyncMessage.h
charlesmchen: I know a lot of people like to do that, but I prefer properties that correspond to state.

src/Devices/OWSVerificationStateSyncMessage.m
charlesmchen: If we did that, I would always be worried about it accidentally getting saved or something.

src/Devices/OWSVerificationStateSyncMessage.m
charlesmchen: Done.

src/Devices/OWSVerificationStateSyncMessage.m
charlesmchen: Done.

protobuf/OWSSignalServiceProtos.proto
charlesmchen: Yup, I am deferring to the property names used by Moxie in his proto schema changes, since we're almost certainly going to adopt those once he's finalized them.

src/Messages/OWSBlockingManager.m
charlesmchen: `ensureLazyInitialization`

src/Messages/OWSBlockingManager.m
charlesmchen: > Do we need to synchronize(self) the call from applicationDidBecomeActive?

Great catch.

> can we make blockedPhoneNumberSet an atomic property since we're potentially reading from it on different threads, and sometimes re-assigning it.

I've made this atomic, but this shouldn't be necessary - it's always accessed within a synchronized block.

src/Messages/OWSIdentityManager.m
charlesmchen: That's about it!

src/Messages/OWSIdentityManager.m
charlesmchen: Done.

src/Messages/OWSIdentityManager.m
charlesmchen: All hail `DDLogErrorAndFail()`

src/Messages/OWSIdentityManager.m
charlesmchen: Heh, we had the same idea.

src/Messages/OWSIdentityManager.m
charlesmchen: I think this is `set` not `sync` - it will trigger a sync but the primary role of this method is to set or update the local verification state.

src/Messages/OWSIdentityManager.m
charlesmchen: Done.

src/Messages/OWSIdentityManager.m
michaelkirk: I'm just trying to find a way to differentiate this log line from the other log line at https://github.com/WhisperSystems/SignalServiceKit/blob/master/src/Messages/OWSIdentityManager.m#L228

```
        DDLogInfo(@"%@ setVerificationState: %@ (%@ -> %@)",
            self.tag,
            recipientId,
            OWSVerificationStateToString(recipientIdentity.verificationState),
            OWSVerificationStateToString(verificationState));
```

PR #253, opened 2017-06-07
Charlesmchen/verification state change messages

charlesmchen:
This PR is now ready for review.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/Interactions/OWSVerificationStateChangeMessage.m
+++ b/src/Messages/Interactions/OWSVerificationStateChangeMessage.m
@@ -15,10 +15,9 @@ NS_ASSUME_NONNULL_BEGIN
                 verificationState:(OWSVerificationState)verificationState
                     isLocalChange:(BOOL)isLocalChange
 {
-    OWSAssert(recipientId.length > 0)
+    OWSAssert(recipientId.length > 0);
 
-        self
-        = [super initWithTimestamp:timestamp inThread:thread messageType:TSInfoMessageVerificationStateChange];
+    self = [super initWithTimestamp:timestamp inThread:thread messageType:TSInfoMessageVerificationStateChange];
     if (!self) {
         return self;
     }
--- a/src/Messages/Interactions/OWSVerificationStateChangeMessage.m
+++ b/src/Messages/Interactions/OWSVerificationStateChangeMessage.m
@@ -1,0 +1,52 @@
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
+
+#import "OWSVerificationStateChangeMessage.h"
+#import "OWSDisappearingMessagesConfiguration.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OWSVerificationStateChangeMessage ()
+
+@property (nonatomic, readonly) OWSVerificationState verificationState;
+@property (nonatomic, readonly) BOOL isLocalChange;
+
+@end
+
+#pragma mark -
+
+@implementation OWSVerificationStateChangeMessage
+
+- (instancetype)initWithTimestamp:(uint64_t)timestamp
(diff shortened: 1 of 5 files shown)


charlesmchen:
This PR is now ready for review.

PTAL @michaelkirk

src/Messages/Interactions/OWSVerificationStateChangeMessage.m
```diff
@@ -0,0 +1,35 @@
...
+
+- (instancetype)initWithTimestamp:(uint64_t)timestamp
+                           thread:(TSThread *)thread
+                      recipientId:(NSString *)recipientId
+                verificationState:(OWSVerificationState)verificationState
+                    isLocalChange:(BOOL)isLocalChange
+{
+    OWSAssert(recipientId.length > 0)
+
+        self
```
charlesmchen: wtf clang-format

src/Messages/OWSIdentityManager.m
```diff
@@ -190,6 +192,14 @@ - (BOOL)saveRemoteIdentity:(NSData *)identityKey recipientId:(NSString *)recipie
             // Cancel any pending verification state sync messages for this recipient.
             [self clearSyncMessageForRecipientId:recipientId];
 
+            // TODO: This may be redundant with the "safety number changes"
+            //       messages.
```
charlesmchen: I suspect we'll _often but not always_ see these right after the "safety number change" messages. I'm wondering about the best way to avoid duplication without missing corner cases.

src/Messages/OWSIdentityManager.m
```diff
@@ -385,18 +407,25 @@ - (void)createIdentityChangeInfoMessageForRecipientId:(NSString *)recipientId
...
     [[TextSecureKitEnv sharedEnv].notificationsManager notifyUserForErrorMessage:errorMessage inThread:contactThread];
 
     for (TSGroupThread *groupThread in [TSGroupThread groupThreadsWithRecipientId:recipientId]) {
-        [[TSErrorMessage nonblockingIdentityChangeInThread:groupThread recipientId:recipientId] save];
+        [messages addObject:[TSErrorMessage nonblockingIdentityChangeInThread:groupThread recipientId:recipientId]];
     }
+
+    [self.storageManager.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *transaction) {
+        for (TSMessage *message in messages) {
+            [message saveWithTransaction:transaction];
```
charlesmchen, line 424: I save everything here (and below) in a single transaction to reduce the number of "yap database modified" notifications.

src/Messages/Interactions/OWSVerificationStateChangeMessage.m
michaelkirk: you forgot a `;` after your assert.

get's me alllll the time.

src/Messages/OWSIdentityManager.m
michaelkirk: WRT a SN change, we should see this only and always when the previous identity was "verified" right (meaning the current one is "no longer verified")?

That seems OK, desirable even.

src/Messages/OWSIdentityManager.m
michaelkirk: That's a good call. Thanks!

michaelkirk (approved):
LGTM

src/Messages/Interactions/OWSVerificationStateChangeMessage.m
charlesmchen: Aha!

src/Messages/OWSIdentityManager.m
charlesmchen: Fair enough. Removing this TODO.

PR #255, opened 2017-06-08
archive not delete

michaelkirk:
Prefer archiving over deleting gives us a little more resiliency in decrypting out of order.

Also removed an overzealous check on the *current* session.

PTAL @charlesmchen 

see also: 

SPK: WhisperSystems/SignalProtocolKit#27
SiOS: https://github.com/WhisperSystems/Signal-iOS/pull/2193

### DIFF
--- a/src/Messages/TSMessagesManager.m
+++ b/src/Messages/TSMessagesManager.m
@@ -336,19 +336,6 @@ NS_ASSUME_NONNULL_BEGIN
         NSString *recipientId = messageEnvelope.source;
         int deviceId = messageEnvelope.sourceDevice;
         dispatch_async([OWSDispatch sessionStoreQueue], ^{
-            if (![storageManager containsSession:recipientId deviceId:deviceId]) {
-                __block TSErrorMessage *errorMessage;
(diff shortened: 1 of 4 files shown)


src/Messages/TSMessagesManager.m
```diff
@@ -336,19 +336,6 @@ - (void)handleSecureMessageAsync:(OWSSignalServiceProtosEnvelope *)messageEnvelo
         NSString *recipientId = messageEnvelope.source;
         int deviceId = messageEnvelope.sourceDevice;
         dispatch_async([OWSDispatch sessionStoreQueue], ^{
-            if (![storageManager containsSession:recipientId deviceId:deviceId]) {
```
michaelkirk, line 339: This method checks if the **current** session is initialized, but what we really care about is if we have **some** (potentially archived) session which can decrypt the method. This behavior is already implemented within session cipher.

So this code is sometimes overzealous, and in the case that it's not overzealous, it's redundant.

charlesmchen (approved):
LGTM

Can we remove `deleteAllSessionsForContact`?

PR #258, opened 2017-06-10
Charlesmchen/refine verification

charlesmchen:
* Don’t update home view sort order in response to dynamic interactions or verification state changes.
* We only want to create change messages in response to user activity, on any of their devices.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSIdentityManager.m
+++ b/src/Messages/OWSIdentityManager.m
@@ -192,12 +192,6 @@ NSString *const kNSNotificationName_IdentityStateDidChange = @"kNSNotificationNa
             // Cancel any pending verification state sync messages for this recipient.
             [self clearSyncMessageForRecipientId:recipientId];
 
-            if (existingIdentity.verificationState != verificationState) {
-                [self saveChangeMessagesForRecipientId:recipientId
-                                     verificationState:verificationState
-                                         isLocalChange:YES];
-            }
-
             [self fireIdentityStateChangeNotification];
 
             return YES;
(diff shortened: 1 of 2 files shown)


src/Messages/OWSIdentityManager.m
```diff
@@ -245,12 +239,12 @@ - (void)setVerificationState:(OWSVerificationState)verificationState
             [self enqueueSyncMessageForVerificationState:verificationState
```
michaelkirk, line 239: small point: since the bool is doing more than determining whether or not we send a sync message, maybe we should call it something else - at the call site it's `isUserInitiatedChange`, maybe that's a better name?

bigger point:

Maybe I'm misreading it, but it seems like with this code we'd no longer print the "change message" when processing a remote verification change received via sync message.

michaelkirk (commented):
provided we're properly handling sync'd verification changes, LGTM.

src/Messages/OWSIdentityManager.m
charlesmchen: > Maybe I'm misreading it, but it seems like with this code we'd no longer print the "change message" when processing a remote verification change received via sync message.

See:
```
            [self saveChangeMessagesForRecipientId:recipientId verificationState:verificationState isLocalChange:NO];
```
in `tryToApplyVerificationStateFromSyncMessage` above.

charlesmchen: `isUserInitiatedChange` Done.

PR #259, opened 2017-06-12
Rework how messages are marked read.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Contacts/TSThread.m
+++ b/src/Contacts/TSThread.m
@@ -240,7 +240,7 @@ NS_ASSUME_NONNULL_BEGIN
     [TSStorageManager.sharedManager.dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction){
         last = [[transaction ext:TSMessageDatabaseViewExtensionName] lastObjectInGroup:self.uniqueId];
     }];
-    return (TSInteraction *)last;
+    return last;
 }
 
 - (TSInteraction *)lastInteractionForInbox
@@ -249,7 +249,7 @@ NS_ASSUME_NONNULL_BEGIN
     [TSStorageManager.sharedManager.dbConnection readWithBlock:^(YapDatabaseReadTransaction *transaction) {
         last = [[transaction ext:TSThreadInboxMessagesDatabaseViewExtensionName] lastObjectInGroup:self.uniqueId];
     }];
-    return (TSInteraction *)last;
+    return last;
 }
 
 - (NSDate *)lastMessageDate {
@@ -277,7 +277,6 @@ NS_ASSUME_NONNULL_BEGIN
         DDLogDebug(@"%@ not showing dynamic interaction in inbox: %@", self.tag, interaction.debugDescription);
         return NO;
     }
-
 
     if ([interaction isKindOfClass:[TSErrorMessage class]]) {
         TSErrorMessage *errorMessage = (TSErrorMessage *)interaction;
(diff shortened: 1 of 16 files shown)


src/Messages/OWSReadTracking.h
```diff
@@ -20,10 +20,8 @@
 - (BOOL)shouldAffectUnreadCounts;
 
 /**
- * Call when the user viewed the message/call on this device. "locally" as opposed to being notified via a read receipt
- * sync message of a remote read.
+ * Used for *responding* to a remote read receipt or in response to user activity.
  */
-- (void)markAsReadLocally;
-- (void)markAsReadLocallyWithTransaction:(YapDatabaseReadWriteTransaction *)transaction;
+- (void)markAsReadWithTransaction:(YapDatabaseReadWriteTransaction *)transaction sendReadReceipt:(BOOL)sendReadReceipt;
```
charlesmchen: Consolidate into a single method that lets caller explicitly specify the desired read receipt behavior.

src/Messages/Interactions/TSIncomingMessage.m
```diff
@@ -115,37 +117,29 @@ - (BOOL)shouldAffectUnreadCounts
...
+    }
 
-- (void)markAsReadWithoutNotificationWithTransaction:(YapDatabaseReadWriteTransaction *)transaction
-{
     DDLogInfo(@"%@ marking as read uniqueId: %@ which has timestamp: %llu", self.tag, self.uniqueId, self.timestamp);
     _read = YES;
     [self saveWithTransaction:transaction];
     [self touchThreadWithTransaction:transaction];
+
+    [OWSDisappearingMessagesJob setExpirationForMessage:self];
```
charlesmchen: Always ping `OWSDisappearingMessagesJob` when any message is marked as read.

charlesmchen:
I just reworked this PR to make two changes that came up in review.

* Make sure we filter messages properly in home view.  
   * Certain kinds of messages shouldn't be shown in home view as the "latest message".  I've added a database view so that we can apply that correct filtering.
   * The exact same filtering should be used when updating the thread's "last message date" which controls thread sorting in home view.
* In addition to marking older messages as read when we receive a read receipt, we should also:
   * Start their "disappearing messages" expiration.
   * Dismiss any local notifications for them.

src/Contacts/TSThread.m
```diff
@@ -259,49 +261,53 @@ - (NSDate *)lastMessageDate {
...
     }
 
+    return YES;
+}
+
+- (void)updateWithLastMessage:(TSInteraction *)lastMessage transaction:(YapDatabaseReadWriteTransaction *)transaction {
+    OWSAssert(lastMessage);
+    OWSAssert(transaction);
+
+    if (![self.class shouldInteractionAppearInInbox:lastMessage]) {
```
michaelkirk, line 306: Good cleanup. 👍

src/Messages/TSCall.m
```diff
@@ -86,22 +86,24 @@ - (BOOL)shouldAffectUnreadCounts
...
-}
 
-- (void)markAsReadLocally
-{
-    [self.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
-        [self markAsReadLocallyWithTransaction:transaction];
-    }];
+    // Ignore sendReadReceipt and updateExpiration; they don't apply to calls.
+    //
+    // TODO: Should we update expiration of calls?
```
michaelkirk, line 106: Good question. Eventually, I think so.

Seems like it will require a little UI work since the current disappearing timer UI is predicated on the bubble-footer context, and has never been applied to a "system message" style interaction.

michaelkirk (commented):
LGTM

src/Messages/TSCall.m
charlesmchen: I raised this in NSG and the feeling was that calls don't and shouldn't expire right now. I vote the other way, but NBD.

PR #260, opened 2017-06-12
Rework how messages are marked read.

charlesmchen:
Slight addendum to https://github.com/WhisperSystems/SignalServiceKit/pull/259

PTAL @michaelkirk

### DIFF
--- a/src/Devices/OWSReadReceiptsProcessor.m
+++ b/src/Devices/OWSReadReceiptsProcessor.m
@@ -90,6 +90,10 @@ NSString *const OWSReadReceiptsProcessorMarkedMessageAsReadNotification =
             // Mark all unread messages in this thread that are older than message specified in the read
             // receipt.
             NSMutableArray<id<OWSReadTracking>> *interactionsToMarkAsRead = [NSMutableArray new];
+
+            // Always mark the message specified by the read receipt as read.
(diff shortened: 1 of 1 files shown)


src/Devices/OWSReadReceiptsProcessor.m
```diff
@@ -90,6 +90,10 @@ - (void)process
             // Mark all unread messages in this thread that are older than message specified in the read
             // receipt.
             NSMutableArray<id<OWSReadTracking>> *interactionsToMarkAsRead = [NSMutableArray new];
+
+            // Always mark the message specified by the read receipt as read.
+            [interactionsToMarkAsRead addObject:message];
```
michaelkirk, line 95: This doesn't seem dangerous or anything, but I'm confused as to how it could fix anything. Can you elaborate on why you made this change?

michaelkirk (commented):
LGTM, but curious.

src/Devices/OWSReadReceiptsProcessor.m
charlesmchen: Yes, this should be redundant but it felt safer to be a little redundant. There's just enough complexity here that I could see future refactoring elsewhere breaking this logic.

PR #261, opened 2017-06-13
Improve logging around incoming messages.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/TSMessagesManager.m
+++ b/src/Messages/TSMessagesManager.m
@@ -605,6 +605,8 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
+    DDLogDebug(@"%@ incoming attachment message: %@", self.tag, createdMessage.debugDescription);
+
     [attachmentsProcessor fetchAttachmentsForMessage:createdMessage
         success:^(TSAttachmentStream *attachmentStream) {
             DDLogDebug(
@@ -892,7 +894,7 @@ NS_ASSUME_NONNULL_BEGIN
                                                                      messageBody:body
                                                                    attachmentIds:attachmentIds
                                                                 expiresInSeconds:dataMessage.expireTimer];
-
+                  DDLogDebug(@"%@ incoming group text message: %@", self.tag, incomingMessage.debugDescription);
                   [incomingMessage saveWithTransaction:transaction];
                   break;
               }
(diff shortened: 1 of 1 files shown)


michaelkirk (commented):
LGTM

src/Messages/TSMessagesManager.m
```diff
@@ -914,12 +916,12 @@ - (TSIncomingMessage *_Nullable)handleReceivedEnvelope:(OWSSignalServiceProtosEn
                                                              messageBody:body
                                                            attachmentIds:attachmentIds
                                                         expiresInSeconds:dataMessage.expireTimer];
+          DDLogDebug(@"%@ incoming 1:1 text message: %@", self.tag, incomingMessage.debugDescription);
+          [incomingMessage saveWithTransaction:transaction];
```
michaelkirk, line 920: Why was this moved? Was there something wrong with the previous logic?

src/Messages/TSMessagesManager.m
```diff
@@ -914,12 +916,12 @@ - (TSIncomingMessage *_Nullable)handleReceivedEnvelope:(OWSSignalServiceProtosEn
...
           thread = cThread;
       }
 
       if (thread && incomingMessage) {
-          [incomingMessage saveWithTransaction:transaction];
```
charlesmchen, line 921: See here.

src/Messages/TSMessagesManager.m
```diff
@@ -892,7 +894,7 @@ - (TSIncomingMessage *_Nullable)handleReceivedEnvelope:(OWSSignalServiceProtosEn
                                                                      messageBody:body
                                                                    attachmentIds:attachmentIds
                                                                 expiresInSeconds:dataMessage.expireTimer];
-
+                  DDLogDebug(@"%@ incoming group text message: %@", self.tag, incomingMessage.debugDescription);
                   [incomingMessage saveWithTransaction:transaction];
```
charlesmchen, line 898: See here.

src/Messages/TSMessagesManager.m
charlesmchen: We were saving messages twice in some cases.  See comments above and below.  We should either save once when the message is created (as I've done here) or once later.  I figured the latter was a little easier to ensure to be correct.

michaelkirk: Got it. Thanks

PR #263, opened 2017-06-13
Charlesmchen/group creation errors

charlesmchen:
* We want to avoid users accidentally creating multiple copies of a group due to partial failures during group creation.  If an error occurs, we navigate the user to the newly created group and show a new "group creation failed" interaction. If tapped, it sends a "group update" in an attempt to fix the group. If any member of the group sends a message, the situation should also be rectified when the "missing" members request a "group update".
* Ensure message sends only succeed or fail once.  We had a bug where fatal errors caused the failure handlers to be called twice.  I've fixed this and added asserts to ensure a problem of this form never re-emerges.

PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -159,6 +159,7 @@ NSUInteger const OWSSendMessageOperationMaxRetries = 4;
 @property (nonatomic, readonly) void (^failureHandler)(NSError *_Nonnull error);
 @property (nonatomic) OWSSendMessageOperationState operationState;
 @property (nonatomic) UIBackgroundTaskIdentifier backgroundTaskIdentifier;
+@property (nonatomic) BOOL hasCompleted;
 
 @end
 
@@ -190,6 +191,13 @@ NSUInteger const OWSSendMessageOperationMaxRetries = 4;
             return;
         }
 
+        // Ensure we call the success or failure handler exactly once.
+        @synchronized(strongSelf)
+        {
+            OWSCAssert(!strongSelf.hasCompleted);
+            strongSelf.hasCompleted = YES;
+        }
+
         [message updateWithMessageState:TSOutgoingMessageStateSentToService];
 
         DDLogDebug(@"%@ succeeded.", strongSelf.tag);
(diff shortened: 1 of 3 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2212

src/Messages/OWSMessageSender.m
```diff
@@ -204,6 +212,13 @@ - (instancetype)initWithMessage:(TSOutgoingMessage *)message
             return;
         }
 
+        // Ensure we call the success or failure handler exactly once.
+        @synchronized(strongSelf)
```
michaelkirk, line 216: No big deal, but we have the existing `markAsComplete` method which has no affect on the new maybe-too-similarly named `hasCompleted` property. 

Instead of adding these lines, did you consider asserting in `markAsComplete` like

```
OWSAssert(self.operationState != OWSSendMessageOperationStateFinished);
```

And avoiding the new `hasCompleted` property all together?

src/Messages/OWSMessageSender.m
```diff
@@ -281,9 +296,14 @@ - (void)tryWithRemainingRetries:(NSUInteger)remainingRetries
 {
     DDLogDebug(@"%@ remainingRetries: %lu", self.tag, (unsigned long)remainingRetries);
 
+    // Use this flag to ensure a given operation only succeeds or fails once.
+    __block BOOL onceFlag = NO;
     RetryableFailureHandler retryableFailureHandler = ^(NSError *_Nonnull error) {
         DDLogInfo(@"%@ Sending failed.", self.tag);
 
+        OWSAssert(!onceFlag);
```
michaelkirk, line 304: Is this redundant with the once-checking that happens in the handlers wrapped in init?

michaelkirk (commented):
LGTM

nice assertions.

src/Messages/OWSMessageSender.m
charlesmchen: Not quite, and:

* There's a lot of complexity in the message sending process. 
* This is core functionality that we want to have high confidence in.
* This flag is cheap and local, so why not.

src/Messages/OWSMessageSender.m
charlesmchen: Not a bad idea.  I'll address this in my "michael's holiday omnibus" PR.

PR #264, opened 2017-06-14
Remove an unnecessary database view.

charlesmchen:
We want to remove unnecessary database views for startup perf.  `TSThreadInboxMessagesDatabaseViewExtensionName` only needs to exist for rare edge cases.  We can just reverse-enumerate with `TSMessageDatabaseViewExtensionName` since 99.9% of the time it will stop after the first object in the collection and it will never have to iterate far.

PTAL @michaelkirk


michaelkirk (commented):
LGTM

PR #265, opened 2017-06-14
Reduce number of database views.

charlesmchen:
First startup times are still slow for users with large numbers of messages, but these changes help (for a user with ~13,000 messages across 4 threads, startup times that were ~80 seconds are now ~50 seconds).  

PTAL @michaelkirk

### DIFF
--- a/src/Storage/TSDatabaseView.h
+++ b/src/Storage/TSDatabaseView.h
@@ -14,17 +14,14 @@ extern NSString *TSSecondaryDevicesGroup;
 
 extern NSString *TSThreadDatabaseViewExtensionName;
 extern NSString *TSMessageDatabaseViewExtensionName;
-extern NSString *TSThreadIncomingMessageDatabaseViewExtensionName;
 extern NSString *TSThreadOutgoingMessageDatabaseViewExtensionName;
 extern NSString *TSUnreadDatabaseViewExtensionName;
 extern NSString *TSUnseenDatabaseViewExtensionName;
-extern NSString *TSDynamicMessagesDatabaseViewExtensionName;
-extern NSString *TSSafetyNumberChangeDatabaseViewExtensionName;
+extern NSString *TSThreadSpecialMessagesDatabaseViewExtensionName;
 extern NSString *TSSecondaryDevicesDatabaseViewExtensionName;
 
 + (BOOL)registerThreadDatabaseView;
 + (BOOL)registerThreadInteractionsDatabaseView;
-+ (BOOL)registerThreadIncomingMessagesDatabaseView;
 + (BOOL)registerThreadOutgoingMessagesDatabaseView;
 
 // Instances of OWSReadTracking for wasRead is NO and shouldAffectUnreadCounts is YES.
@@ -37,8 +34,7 @@ extern NSString *TSSecondaryDevicesDatabaseViewExtensionName;
 // Instances of OWSReadTracking for wasRead is NO.
 + (BOOL)registerUnseenDatabaseView;
 
-+ (BOOL)registerDynamicMessagesDatabaseView;
-+ (BOOL)registerSafetyNumberChangeDatabaseView;
++ (BOOL)registerThreadSpecialMessagesDatabaseView;
 + (void)asyncRegisterSecondaryDevicesDatabaseView;
 
 @end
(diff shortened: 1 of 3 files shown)


src/Storage/TSDatabaseView.h
```diff
@@ -14,17 +14,14 @@ extern NSString *TSSecondaryDevicesGroup;
 
 extern NSString *TSThreadDatabaseViewExtensionName;
 extern NSString *TSMessageDatabaseViewExtensionName;
-extern NSString *TSThreadIncomingMessageDatabaseViewExtensionName;
```
charlesmchen, line 17: I just iterate in TSMessageDatabaseViewExtensionName now instead of using this view.

src/Storage/TSDatabaseView.h
```diff
@@ -14,17 +14,14 @@ extern NSString *TSSecondaryDevicesGroup;
...
 extern NSString *TSThreadOutgoingMessageDatabaseViewExtensionName;
 extern NSString *TSUnreadDatabaseViewExtensionName;
 extern NSString *TSUnseenDatabaseViewExtensionName;
-extern NSString *TSDynamicMessagesDatabaseViewExtensionName;
-extern NSString *TSSafetyNumberChangeDatabaseViewExtensionName;
+extern NSString *TSThreadSpecialMessagesDatabaseViewExtensionName;
```
charlesmchen, line 20: I've combined these two database views.

src/Storage/TSDatabaseView.m
```diff
@@ -20,12 +20,10 @@
 
 NSString *TSThreadDatabaseViewExtensionName  = @"TSThreadDatabaseViewExtensionName";
 NSString *TSMessageDatabaseViewExtensionName = @"TSMessageDatabaseViewExtensionName";
-NSString *TSThreadIncomingMessageDatabaseViewExtensionName = @"TSThreadOutgoingMessageDatabaseViewExtensionName";
```
charlesmchen, line 23: Note that this constant was wrong - it was a collision with the next database view.  Removing this still improves startup times, but I don't understand how we didn't notice incorrect behavior.

src/Storage/TSDatabaseView.m
```diff
@@ -127,37 +110,22 @@ + (BOOL)registerSafetyNumberChangeDatabaseView
         return nil;
     }];
 
-    return [self registerMessageDatabaseViewWithName:TSSafetyNumberChangeDatabaseViewExtensionName
+    return [self registerMessageDatabaseViewWithName:TSThreadSpecialMessagesDatabaseViewExtensionName
                                         viewGrouping:viewGrouping
-                                             version:@"2"];
+                                             version:@"1"];
```
charlesmchen: I'm bumping this version number due to the collision noted above.

src/Storage/TSDatabaseView.m
```diff
@@ -127,37 +110,22 @@ + (BOOL)registerSafetyNumberChangeDatabaseView
...
 }
 
 + (BOOL)registerThreadInteractionsDatabaseView
 {
     YapDatabaseViewGrouping *viewGrouping = [YapDatabaseViewGrouping withObjectBlock:^NSString *(
         YapDatabaseReadTransaction *transaction, NSString *collection, NSString *key, id object) {
-        if ([object isKindOfClass:[TSInteraction class]]) {
```
charlesmchen, line 139: Just assert this, so that we can avoid this check in production.

src/Storage/TSDatabaseView.m
```diff
@@ -174,7 +142,7 @@ + (BOOL)registerThreadOutgoingMessagesDatabaseView
 
     return [self registerMessageDatabaseViewWithName:TSThreadOutgoingMessageDatabaseViewExtensionName
                                         viewGrouping:viewGrouping
-                                             version:@"1"];
+                                             version:@"2"];
```
charlesmchen, line 145: I'm bumping this version number due to the collision noted above.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2215

charlesmchen:
@michaelkirk It's worth noting that the first startup time measurements mentioned above are on our worst-case device, the iPhone 5c.

src/Storage/TSDatabaseView.m
michaelkirk: I think this only affected whether or not we showed the block offer, right?

michaelkirk:
LGTM

src/Storage/TSDatabaseView.m
charlesmchen: Yup.

charlesmchen: Just to be clear, this will also eventually affect the "share profile" behavior.  But this naming issue is resolved.

PR #266, opened 2017-06-14
Remove “block on safety number changes” setting in preferences.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2217

michaelkirk (commented):
LGTM

PR #267, opened 2017-06-15
Charlesmchen/database views vs startup time

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSDisappearingMessagesFinder.h
+++ b/src/Messages/OWSDisappearingMessagesFinder.h
@@ -1,38 +1,39 @@
-//  Created by Michael Kirk on 9/23/16.
-//  Copyright © 2016 Open Whisper Systems. All rights reserved.
+//
+//  Copyright (c) 2017 Open Whisper Systems. All rights reserved.
+//
 
 NS_ASSUME_NONNULL_BEGIN
 
 @class TSStorageManager;
 @class TSMessage;
 @class TSThread;
+@class YapDatabaseReadTransaction;
+@class YapDatabaseReadWriteTransaction;
 
 @interface OWSDisappearingMessagesFinder : NSObject
 
-- (instancetype)init NS_UNAVAILABLE;
-- (instancetype)initWithStorageManager:(TSStorageManager *)storageManager NS_DESIGNATED_INITIALIZER;
-
-+ (instancetype)defaultInstance;
-
-- (void)enumerateExpiredMessagesWithBlock:(void (^_Nonnull)(TSMessage *message))block;
-- (void)enumerateUnstartedExpiringMessagesInThread:(TSThread *)thread block:(void (^_Nonnull)(TSMessage *message))block;
+- (void)enumerateExpiredMessagesWithBlock:(void (^_Nonnull)(TSMessage *message))block
+                              transaction:(YapDatabaseReadWriteTransaction *_Nonnull)transaction;
+- (void)enumerateUnstartedExpiringMessagesInThread:(TSThread *)thread
+                                             block:(void (^_Nonnull)(TSMessage *message))block
+                                       transaction:(YapDatabaseReadWriteTransaction *_Nonnull)transaction;
 
 /**
  * @return
  *   uint64_t millisecond timestamp wrapped in a number. Retrieve with `unsignedLongLongvalue`.
  *   or nil if there are no upcoming expired messages
  */
-- (nullable NSNumber *)nextExpirationTimestamp;
+- (nullable NSNumber *)nextExpirationTimestampWithTransaction:(YapDatabaseReadTransaction *_Nonnull)transaction;
 
 /**
  * Database extensions required for class to work.
  */
-- (void)asyncRegisterDatabaseExtensions;
++ (void)asyncRegisterDatabaseExtensions:(TSStorageManager *)storageManager;
 
 /**
  * Only use the sync version for testing, generally we'll want to register extensions async
  */
-- (void)blockingRegisterDatabaseExtensions;
++ (void)blockingRegisterDatabaseExtensions:(TSStorageManager *)storageManager;
 
 @end
 
--- a/src/Messages/OWSDisappearingMessagesFinder.h
+++ b/src/Messages/OWSDisappearingMessagesFinder.h
@@ -8,15 +8,14 @@ NS_ASSUME_NONNULL_BEGIN
 @class TSMessage;
 @class TSThread;
 @class YapDatabaseReadTransaction;
-@class YapDatabaseReadWriteTransaction;
 
 @interface OWSDisappearingMessagesFinder : NSObject
 
 - (void)enumerateExpiredMessagesWithBlock:(void (^_Nonnull)(TSMessage *message))block
-                              transaction:(YapDatabaseReadWriteTransaction *_Nonnull)transaction;
+                              transaction:(YapDatabaseReadTransaction *)transaction;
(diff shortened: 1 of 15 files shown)


charlesmchen:
* Use transactions in the jobs.
Avoid crashing on startup due to database view creation.
* Substitute unread view for unseen view until unseen view is ready.
* Register as many views as possible async.
* Perform blocking, safe migrations before async registration of views.

charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2219

src/Messages/OWSDisappearingMessagesFinder.m
```diff
@@ -17,60 +18,33 @@
...
-    static OWSDisappearingMessagesFinder *defaultInstance = nil;
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        defaultInstance = [[self alloc] initWithStorageManager:[TSStorageManager sharedManager]];
-    });
-    return defaultInstance;
-}
-
 - (NSArray<NSString *> *)fetchUnstartedExpiringMessageIdsInThread:(TSThread *)thread
+                                                      transaction:(YapDatabaseReadWriteTransaction *_Nonnull)transaction
```
michaelkirk: nit: I think all the `YapDatabaseReadWriteTransaction` parameters in this file can be `YapDatabaseReadTransaction`'s.

You can safely pass in a ReadWrite transaction (which subclasses Read transaction).

But we're not using the write API here, and if we use these methods elsewhere it doesn't make sense to require a writeable transaction be created.

src/Messages/OWSDisappearingMessagesFinder.h
```diff
@@ -1,38 +1,39 @@
...
 
-- (instancetype)init NS_UNAVAILABLE;
-- (instancetype)initWithStorageManager:(TSStorageManager *)storageManager NS_DESIGNATED_INITIALIZER;
-
-+ (instancetype)defaultInstance;
-
-- (void)enumerateExpiredMessagesWithBlock:(void (^_Nonnull)(TSMessage *message))block;
-- (void)enumerateUnstartedExpiringMessagesInThread:(TSThread *)thread block:(void (^_Nonnull)(TSMessage *message))block;
+- (void)enumerateExpiredMessagesWithBlock:(void (^_Nonnull)(TSMessage *message))block
+                              transaction:(YapDatabaseReadWriteTransaction *_Nonnull)transaction;
```
michaelkirk: style nit: `(nonnull YapDatabaseReadWriteTransaction *)`

but per another comment, I think we want: `(nonnull YapDatabaseReadTransaction *)`

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -84,16 +88,20 @@ - (void)run
     uint64_t now = [NSDate ows_millisecondTimeStamp];
 
     __block uint expirationCount = 0;
-    [self.disappearingMessagesFinder enumerateExpiredMessagesWithBlock:^(TSMessage *message) {
-        // sanity check
-        if (message.expiresAt > now) {
-            DDLogError(@"%@ Refusing to remove message which doesn't expire until: %lld", self.tag, message.expiresAt);
-            return;
+    [self.storageManager.dbConnection readWriteWithBlock:^(YapDatabaseReadWriteTransaction *_Nonnull transaction) {
```
michaelkirk: I'm starting to feel that having a global dbConnection on storageManager might be an anti-pattern.

Previously the finder had it's own dbConnection (via `newDatabaseConnection`) which means it's operations had their own serial queue. Since that's been removed, I think this job should have it's own dbConnection if we don't want it to interfere with(/be interfered with) other things that share the storageManager.dbConnection's serial queue.

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -18,6 +18,9 @@
 @interface OWSDisappearingMessagesJob ()
 
 // This property should only be accessed on the serialQueue.
+@property (nonatomic, readonly) TSStorageManager *storageManager;
```
michaelkirk: re: comment, where is this serialQueue access being enforced? 

Also not sure why it would be necessary to constrain access to the dedicated serial queue since it's readonly and assigned in init.

src/Storage/TSDatabaseView.h
```diff
@@ -5,36 +5,53 @@
...
++ (void)asyncRegisterThreadSpecialMessagesDatabaseView;
 
-+ (BOOL)registerThreadSpecialMessagesDatabaseView;
 + (void)asyncRegisterSecondaryDevicesDatabaseView;
 
+// Returns the "unseen" database view if it is ready;
+// otherwise it returns the "unread" database view.
++ (id)unseenDatabaseViewExtension:(YapDatabaseReadTransaction *)transaction;
+
+// NOTE: It is not safe to call this method until hasPendingViewRegistrations is YES.
```
michaelkirk: comment typo I think:
```
- // NOTE: It is not safe to call this method until hasPendingViewRegistrations is YES.
+ // NOTE: It is not safe to call this method __while__ hasPendingViewRegistrations is YES.
```

src/Messages/OWSDisappearingMessagesJob.m
```diff
@@ -142,15 +154,19 @@ - (void)setExpirationForMessage:(TSMessage *)message
     [self setExpirationForMessage:message expirationStartedAt:[NSDate ows_millisecondTimeStamp]];
 }
 
-+ (void)setExpirationForMessage:(TSMessage *)message expirationStartedAt:(uint64_t)expirationStartedAt
+- (void)setExpirationForMessage:(TSMessage *)message expirationStartedAt:(uint64_t)expirationStartedAt
 {
-    dispatch_async(self.serialQueue, ^{
```
michaelkirk: Why is this no longer async? And why do we no longer want it on a serialized queue?

michaelkirk (commented):
Had a couple of questions...

src/Messages/OWSDisappearingMessagesFinder.m
charlesmchen: Done.

src/Messages/OWSDisappearingMessagesFinder.h
charlesmchen: Assume non-null, so removing altogether.

src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: I've removed these comments but added others indicating which methods should only be called on the `serialQueue`.

src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: > I think this job should have it's own dbConnection

Good idea, done.

charlesmchen: I also moved `ensureDynamicInteractionsForThread` to the message view's `uiDatabaseConnection`.

src/Storage/TSDatabaseView.h
charlesmchen: Ah thank you.

src/Messages/OWSDisappearingMessagesJob.m
charlesmchen: This is a misleading diff, I think. The class method got removed by accident. I've restored it.

charlesmchen:
PTAL @michaelkirk

michaelkirk:
LGTM

PR #272, opened 2017-06-16
Charlesmchen/holiday code review omnibus

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -195,12 +195,7 @@ NSUInteger const OWSSendMessageOperationMaxRetries = 4;
         DDLogDebug(@"%@ succeeded.", strongSelf.tag);
         aSuccessHandler();
 
-        // Ensure we call the success or failure handler exactly once.
-        @synchronized(strongSelf)
-        {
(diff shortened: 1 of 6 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2226

src/Messages/OWSMessageSender.m
```diff
@@ -212,18 +210,17 @@ - (instancetype)initWithMessage:(TSOutgoingMessage *)message
...
-        }
-
         [strongSelf.message updateWithSendingError:error];
 
         DDLogDebug(@"%@ failed with error: %@", strongSelf.tag, error);
         aFailureHandler(error);
-        [strongSelf markAsComplete];
+
+        // Ensure we call the success or failure handler exactly once.
+        @synchronized(strongSelf)
```
michaelkirk: nit: I think we could just do the check *in* `markAsComplete` and dedupe the two asserts blocks.

michaelkirk (approved):
LGTM

src/Messages/OWSMessageSender.m
charlesmchen: Done.

PR #273, opened 2017-06-19
message sending must be on main thread

michaelkirk:
include description for verification sync messages

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #274, opened 2017-06-19
append/remove key type as necessary to fix verification syncing

michaelkirk:
PTAL @charlesmchen 

Note that I had to locally toggle the sync feature-flag to test this.

It's kind of hairy, but I was trying to move as little code as possible before the release.

Longer term we should migrate the db to simplify the code paths.


charlesmchen (approved):
LGTM

PR #276, opened 2017-06-20
Refine observation of async registration completion.

charlesmchen:
PTAL @michaelkirk

### DIFF
--- a/src/Storage/TSDatabaseView.h
+++ b/src/Storage/TSDatabaseView.h
@@ -54,4 +54,7 @@ extern NSString *const TSSecondaryDevicesDatabaseViewExtensionName;
 // NOTE: It is not safe to call this method while hasPendingViewRegistrations is YES.
 + (id)threadSpecialMessagesDatabaseView:(YapDatabaseReadTransaction *)transaction;
 
+// This method should be called _after_ all async database registrations have been started.
++ (void)asyncRegistrationCompletion;
+
 @end
--- a/src/Storage/TSStorageManager.m
+++ b/src/Storage/TSStorageManager.m
@@ -225,6 +225,11 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
     OWSFailedAttachmentDownloadsJob *failedAttachmentDownloadsMessagesJob =
         [[OWSFailedAttachmentDownloadsJob alloc] initWithStorageManager:self];
     [failedAttachmentDownloadsMessagesJob asyncRegisterDatabaseExtensions];
+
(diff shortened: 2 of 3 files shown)


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2248

src/Storage/TSStorageManager.m
```diff
@@ -225,6 +225,11 @@ - (void)setupDatabaseWithSafeBlockingMigrations:(void (^_Nonnull)())safeBlocking
     OWSFailedAttachmentDownloadsJob *failedAttachmentDownloadsMessagesJob =
         [[OWSFailedAttachmentDownloadsJob alloc] initWithStorageManager:self];
     [failedAttachmentDownloadsMessagesJob asyncRegisterDatabaseExtensions];
+
+    // NOTE: [TSDatabaseView asyncRegistrationCompletion] ensures that
+    // kNSNotificationName_DatabaseViewRegistrationComplete is not fired until all
+    // of the async registrations are complete.
+    [TSDatabaseView asyncRegistrationCompletion];
```
charlesmchen, line 232: Rather than just to get all of the "async database registrations" participating in the old completion observation mechanism, I just went for something similar - using an async write to block on completion.

src/Storage/TSDatabaseView.h
```diff
@@ -54,4 +54,7 @@ extern NSString *const TSSecondaryDevicesDatabaseViewExtensionName;
 // NOTE: It is not safe to call this method while hasPendingViewRegistrations is YES.
 + (id)threadSpecialMessagesDatabaseView:(YapDatabaseReadTransaction *)transaction;
 
+// This method should be called _after_ all async database registrations have been started.
++ (void)asyncRegistrationCompletion;
```
michaelkirk, line 58: I **think** this is a comment typo?

```
-// This method should be called _after_ all async database registrations have been started.
+// This method should be called _after_ all async database registrations have been __completed__.
```

michaelkirk (approved):
That's clever. LGTM

src/Storage/TSDatabaseView.h
charlesmchen: I guess it is confusing.  `TSStorageManager` kicks of the async registrations but doesn't wait  for them to complete, hence started.  I'm not sure how how to make these less clear, but IMO `completed` is more misleading.

michaelkirk: Ah, I did get confused. Maybe `notifyUponAsyncRegistrationCompletion`

PR #280, opened 2017-06-21
no need to send sync messages when only 1 device

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -1089,10 +1089,16 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 - (void)handleMessageSentLocally:(TSOutgoingMessage *)message
 {
     if (message.shouldSyncTranscript) {
-        // TODO: I suspect we shouldn't optimistically set hasSyncedTranscript.
-        //       We could set this in a success handler for [sendSyncTranscriptForMessage:].
-        [message updateWithHasSyncedTranscript:YES];
-        [self sendSyncTranscriptForMessage:message];
+        __block BOOL hasSecondaryDevices = NO;
+        [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
+        if (hasSecondaryDevices) {
+            // TODO: I suspect we shouldn't optimistically set hasSyncedTranscript.
+            //       We could set this in a success handler for [sendSyncTranscriptForMessage:].
+            [message updateWithHasSyncedTranscript:YES];
+            [self sendSyncTranscriptForMessage:message];
+        }
     }
 
     [OWSDisappearingMessagesJob setExpirationForMessage:message];
--- a/src/Messages/OWSMessageSender.m
+++ b/src/Messages/OWSMessageSender.m
@@ -594,6 +594,19 @@ NSString *const OWSMessageSenderRateLimitedException = @"RateLimitedException";
 {
     TSThread *thread = message.thread;
 
+#ifdef DEBUG
+    // Assert that we're only sending sync messages when necessary
+    if ([message isKindOfClass:[OWSOutgoingSyncMessage class]]) {
+        __block BOOL hasSecondaryDevices = NO;
+        [self.storageManager.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
+        if (!hasSecondaryDevices) {
+            OWSFail(@"Sending sync message when we have no devices to sync to.");
+        }
+    }
+#endif
+
     dispatch_async([OWSDispatch sendingQueue], ^{
         if ([thread isKindOfClass:[TSGroupThread class]]) {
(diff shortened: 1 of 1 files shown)


charlesmchen (approved):
The service thanks you. LGTM

src/Messages/OWSMessageSender.m
```diff
@@ -1089,10 +1089,16 @@ - (void)handleMismatchedDevices:(NSDictionary *)dictionary
...
-        [self sendSyncTranscriptForMessage:message];
+        __block BOOL hasSecondaryDevices = NO;
+        [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
+        if (hasSecondaryDevices) {
+            // TODO: I suspect we shouldn't optimistically set hasSyncedTranscript.
+            //       We could set this in a success handler for [sendSyncTranscriptForMessage:].
+            [message updateWithHasSyncedTranscript:YES];
+            [self sendSyncTranscriptForMessage:message];
```
charlesmchen, line 1113: Could we generalize this in `[OWSMessageSender deliverMessage:...]`?

charlesmchen: Somewhere around here:
```
            NSString *recipientContactId = [message isKindOfClass:[OWSOutgoingSyncMessage class]]
                ? self.storageManager.localNumber
                : contactThread.contactIdentifier;

```

michaelkirk:
> Could we generalize this in [OWSMessageSender deliverMessage:...]?
  
Good idea.

I pushed up a another commit which makes debug asserts. I think it will be better to just not build these messages at all.

https://github.com/WhisperSystems/SignalServiceKit/pull/280/commits/2737845c25ad501232fc5495f49773dde7c4b0fc

Can you take another look @charlesmchen?

src/Messages/OWSMessageSender.m
```diff
@@ -594,6 +594,19 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
...
+#ifdef DEBUG
+    // Assert that we're only sending sync messages when necessary
+    if ([message isKindOfClass:[OWSOutgoingSyncMessage class]]) {
+        __block BOOL hasSecondaryDevices = NO;
+        [self.storageManager.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
+        if (!hasSecondaryDevices) {
+            OWSFail(@"Sending sync message when we have no devices to sync to.");
+        }
```
charlesmchen, line 606: Can't we return too so that we get the right behavior in prod?

src/Messages/OWSMessageSender.m
```diff
@@ -594,6 +594,19 @@ - (void)deliverMessage:(TSOutgoingMessage *)message
 {
     TSThread *thread = message.thread;
 
+#ifdef DEBUG
+    // Assert that we're only sending sync messages when necessary
+    if ([message isKindOfClass:[OWSOutgoingSyncMessage class]]) {
+        __block BOOL hasSecondaryDevices = NO;
+        [self.storageManager.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
```
charlesmchen, line 601: Here you use self.storageManager.dbConnection but below you use self.dbConnection.

src/Messages/OWSMessageSender.m
```diff
@@ -1089,10 +1102,16 @@ - (void)handleMismatchedDevices:(NSDictionary *)dictionary
...
 {
     if (message.shouldSyncTranscript) {
-        // TODO: I suspect we shouldn't optimistically set hasSyncedTranscript.
-        //       We could set this in a success handler for [sendSyncTranscriptForMessage:].
-        [message updateWithHasSyncedTranscript:YES];
-        [self sendSyncTranscriptForMessage:message];
+        __block BOOL hasSecondaryDevices = NO;
+        [self.dbConnection readWithBlock:^(YapDatabaseReadTransaction *_Nonnull transaction) {
+            hasSecondaryDevices = [OWSDevice hasSecondaryDevicesWithTransaction:transaction];
+        }];
```
charlesmchen, line 1108: seems like we could pull this into `OWSDevice` like this: `hasSecondaryDevicesWithConnection:`

charlesmchen (approved):
LGTM Just suggestions.

src/Messages/OWSMessageSender.m
michaelkirk: good catch!

src/Messages/OWSMessageSender.m
michaelkirk: Yeah, there's a lot of boilerplate here, but I'm learning to avoid creating transactions in model instance methods, as they often lead to deadlock down the road when you forget they create their own transaction.

src/Messages/OWSMessageSender.m
michaelkirk: I started to go down this road, then realized that there's a race condition where a newly linked device immediately sends a group/contact sync request, before this device even knows about the device.

There's also an issue where it's possible to *not* persist the devices if the user leaves the device linking UI *just* as it completes.

So I'm going to put this on hold until the next release when we have more time to test it.

PR #281, opened 2017-06-22
[SSK] Reduce chattiness of logs; increase log file sizes.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/Signal-iOS/pull/2260


michaelkirk (approved):
LGTM

PR #282, opened 2017-06-22
Add creation timestamp to attachment streams.

charlesmchen:
We're going to use this in the future when cleaning up orphan timestamps.  It's important to get it into v2.13.0.0 so that we can piggyback on the "save all TSAttachmentStream instances" migration.

PTAL @michaelkirk


michaelkirk (approved):
LGTM

PR #283, opened 2017-06-22
Archive sessions upon identity change.

michaelkirk:
Session clearing must be in IdentityKeyStore implementation

Trying to do it in the protocol will not work with multiple devices,
because we'll only archive the session of the first sending device.

See also: https://github.com/WhisperSystems/SignalProtocolKit/pull/29

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #284, opened 2017-06-22
Fix broken tests.

charlesmchen:
PTAL @michaelkirk


charlesmchen:
See: https://github.com/WhisperSystems/Signal-iOS/pull/2266

michaelkirk (commented):
There are a lot of singleton asserts in the code base. I think we should solve this within the assert so we don't have to add something like this to every class.

e.g. see my existing PR: https://github.com/WhisperSystems/SignalServiceKit/pull/244

Or convert them all to protocols 😬

charlesmchen:
Interesting, let's talk on your PR.

michaelkirk:
Closing as I think #244 took care of everything. LMK if I missed something.

PR #289, opened 2017-07-01
FIX: verifiying unregistered user prints "no longer registered" error on

michaelkirk:
every launch

This happens if you ever try to mark an unregistered user as (un)/verified.

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #290, opened 2017-07-03
call type for declined call

michaelkirk:
PTAL @charlesmchen

### DIFF
--- a/src/Messages/TSCall.m
+++ b/src/Messages/TSCall.m
@@ -79,9 +79,6 @@ NSUInteger TSCallCurrentSchemaVersion = 1;
         case RPRecentCallTypeIncomingDeclined:
             return NSLocalizedString(@"INCOMING_DECLINED_CALL",
                 @"info message recorded in conversation history when local user declined a call");
-        default:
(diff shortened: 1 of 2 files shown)


src/Messages/TSCall.m
```diff
@@ -76,6 +76,12 @@ - (NSString *)description {
             return NSLocalizedString(@"INCOMING_INCOMPLETE_CALL", @"");
         case RPRecentCallTypeMissedBecauseOfChangedIdentity:
             return NSLocalizedString(@"INFO_MESSAGE_MISSED_CALL_DUE_TO_CHANGED_IDENITY", @"info message text shown in conversation view");
+        case RPRecentCallTypeIncomingDeclined:
+            return NSLocalizedString(@"INCOMING_DECLINED_CALL",
+                @"info message recorded in conversation history when local user declined a call");
+        default:
```
charlesmchen: I'd rather handle the `default` outside the switch statement.  If you add a `default` case, the compiler no longer warns you when you having a messing enum value, which is very useful.

PR #291, opened 2017-07-04
Run orphan cleanup on startup.

charlesmchen:
PTAL @michaelkirk 

See: https://github.com/WhisperSystems/Signal-iOS/pull/2291


src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -9,82 +11,209 @@
...
 
-    // Remove any lingering attachments
-    for (NSString *path in [self orphanedFilePaths]) {
-        DDLogWarn(@"Removing orphaned file attachment at path: %@", path);
+// This method finds and optionally cleans up:
+//
+// * Orphan messages (with no thread).
+// * Orphan attachments (with no message).
+// * Orphan attachment files (with no attachment).
+// * Missing attachment files (cannot be cleaned up).
```
michaelkirk, line 43: > // * Missing attachment files (cannot be cleaned up).

Do you mean (e.g.) TSAttachmentStream's with no corresponding on disk attachment?

by "(cannot be cleaned up)" I assume you mean we can detect it, but we cannot delete the AttachmentStream? Can you expound on the rationale in the comment?

src/Storage/OWSOrphanedDataCleaner.m
```diff
@@ -9,82 +11,209 @@
...
+
+    DDLogDebug(@"orphan attachmentIds: %zd", orphanAttachmentIds.count);
+    DDLogDebug(@"missing attachmentIds: %zd", missingAttachmentIds.count);
+    DDLogDebug(@"orphan interactions: %zd", orphanInteractionIds.count);
+
+    // We need to avoid cleaning up new attachments and files that are still in the process of
+    // being created/written, so we don't clean up anything recent.
+    const NSTimeInterval kMinimumOrphanAge = 15 * 60.f;
+
+    if (shouldCleanup) {
```
michaelkirk: style nit: invert, and early return here avoids a lengthy conditional block, which improves readability and might help save us from a catastrophic mismatched brace in the future.

    if (!shouldCleanup) {
        return;
    }

michaelkirk (approved):
LGTM

michaelkirk:
@charlesmchen  Looks like there are some new test failures. Can you fix them before merging?

src/Storage/OWSOrphanedDataCleaner.m
charlesmchen: Right. Done.

src/Storage/OWSOrphanedDataCleaner.m
charlesmchen: Done.

PR #295, opened 2017-07-06
fix tests

michaelkirk:
I should've run that reverse integration through PR so I could see the tests fail...

PTAL @charlesmchen


charlesmchen (approved):
LGTM

PR #297, opened 2017-07-10
remove unhelpful logging

michaelkirk:
This is open for line item veto

PTAL @charlesmchen


src/Contacts/PhoneNumber.m
```diff
@@ -21,11 +21,6 @@ + (PhoneNumber *)phoneNumberFromText:(NSString *)text andRegion:(NSString *)regi
     NSError *parseError   = nil;
     NBPhoneNumber *number = [phoneUtil parse:text defaultRegion:regionCode error:&parseError];
 
-    if (parseError) {
-        DDLogDebug(@"Issue while parsing number: %@", [parseError description]);
```
charlesmchen: Don't we still want to early exit in this case?

michaelkirk: definitely yes. Good catch.

michaelkirk:
PTAL @charlesmchen

charlesmchen (approved):
LGTM

PR #298, opened 2017-07-10
Fix persist view for upgrade scenarios

michaelkirk:
PTAL @charlesmchen


charlesmchen (approved):
LGTM

Question: Would it be sufficient to bump the version number here:

```
    YapDatabaseView *databaseView =
        [[YapDatabaseView alloc] initWithGrouping:viewGrouping sorting:viewSorting versionTag:@"1" options:options];

    [[TSStorageManager sharedManager].database registerExtension:databaseView
                                                        withName:TSThreadDatabaseViewExtensionName];
```

michaelkirk:
> Question: Would it be sufficient to bump the version number here:

Sadly no - I tried that.

PR #299, opened 2017-07-11
update ci gems

michaelkirk:
Maybe this will help our flaky tests.

"error 65" can be caused by lots of things - but in my experience it's usually a simulator boot timeout on travis. 

PTAL @charlesmchen


charlesmchen (approved):
LGTM

